PR #91, opened 2019-08-06
Fix/spikeapps

danielgmu:


danielgmu:
Added demo file at root directory, should we have a demo folder?

stevevanhooser:
I made a couple changes (to tabs and making a demo folder as you suggest). Now I will do a new pull request.

PR #132, opened 2020-06-01
Sessiontable1

stevevanhooser:


stevevanhooser:
Adds session table, simplifies global variable to a simple "ndi" structure

PR #269, opened 2024-05-12
Add a DaqSystemConfiguration class and json configuration files for initializing DAQ systems

ehennestad:
This PR should make it easier to pre-configure and re-use pre-configured DAQ systems.

Some notes: 
1) We should think if the DAQSystemConfiguration class needs more properties and if their types are correctly set

2) I updated all the daq.system.\<lab\> functions to load DAQ SYSTEM configurations from a json in the ndi_common/daq_systems folder. I am not able to verify/test this for all the labs. We need to think about how to verify this properly


stevevanhooser:
Hi @ehennestad : Looking good, except one test fails. 

`ndi.test.database.core.test_ndi_daq_documents`

It is looking for `ndi.setup.daq.system.vhlab`, which I gather is deprecated. Can you look into what the replacement code should be?

My normal test suite is to run

`ndi.test.testsuite.ndi_testsuite`

It checks a whole bunch of things..takes about 1-2 minutes to run.

Thanks
Steve

stevevanhooser:
I also use the following function to open experiments from our lab for the first time:

`S = ndi.setup.vhlab(reference, pathname)`

I imagine fixing the above will allow that function to work, too, right?

ehennestad:
`S = ndi.setup.vhlab(reference, pathname)` should work.

I updated the test in question and it runs for me

ehennestad:
Can potentially also keep the `ndi.setup.daq.system.vhlab` and similar functions if preferred

stevevanhooser:
Ah yes, I see, just that function is unavailable, the top-level function is still there. Thanks

PR #272, opened 2024-06-04
Add routines for installing dependency from FileExchange

ehennestad:
I added a function for installing dependencies from FileExchange. It should install zipped packages and matlab toolboxes, I am not aware of other cases.

For a FEX dependency, I went for the pattern fex//:<fex_submission_uuid>. Unfortunately, the uuid does not say anything about the name of the package, so it gets very opaque.


stevevanhooser:
Thanks @ehennestad !

PR #284, opened 2024-07-02
Update ndi install

ehennestad:
- Provide option to install dependencies from locally checked out branch
- Restore the user's path when running an update


stevevanhooser (commented):
looks good, thanks

stevevanhooser (approved):
looks good thanks

PR #290, opened 2024-10-06
Improve performance when finding filegroups

ehennestad:
This change uses memoization to improve the performance when finding file groups, as this function might get called many times with the same inputs


stevevanhooser (approved):
Thanks! Nice change

PR #292, opened 2024-10-14
Remove ndi.globals

ehennestad:


stevevanhooser:
Hi @ehennestad -

I am fixing a few small bugs (for example, `did_Init.m` no longer exists so it can't be used to find did), but there's one where I'm unsure about your preferences.

I have two `NDIcalc-*-matlab` directories. The following is failing in the initialization of the `ndi.common.PathConstants`:


```
function mustUpdateDidGlobals(value, key)
    ndi.common.assertDIDInstalled()
    definitionsMap = did.common.PathConstants.definitions;
    if ~isKey(definitionsMap, key)
        if ~isempty(value)
            definitionsMap(key) = value; %#ok<NASGU> This is a handle object
        else
            if strcmp(key, "$NDIDOCUMENTPATH")
                error('Could not update DID globals')
            else
                %pass
            end
        end
    end
end
```

The reason it fails is because the value is a cell array of character arrays, whereas key only accepts character arrays. 

I could see many ... uh, paths ... forward. 

1. We could make the definition keyType set to `any`
2. We could make the character strings like a path with `:` separators.

Or other ideas. Any preference?

stevevanhooser:
Provisionally, I am proceeding with it set to "any"

ehennestad:
Do you have these ndiCalc directories as repositories? I think I need to see it to understand what it looks like

ehennestad:
> The reason it fails is because the value is a cell array of character arrays, whereas key only accepts character arrays.

I tried to run this with to ndiCalc projects on path and it worked, so I did not manage to reproduce the error.

I am curious about how this works if the placeholder i.e `$NDICALCDOCUMENT` is replaced with the location, should all of the locations in the cell array go into the new string?

stevevanhooser:
Interesting. I'm working on a fix, hopefully by end of night. 

The error arises because of the following:

```
d = ndi.fun.find_calc_directories'

d =

  2×1 cell array

    {'/Users/vanhoosr/Documents/MATLAB/tools/NDIcalc-vis-matlab'     }
    {'/Users/vanhoosr/Documents/MATLAB/tools/NDIcalc-visbayes-matlab'}
```

When it tries to assign this to Containers.Map (at least in 2024a) the Containers.Map keyType has been set automatically to 'char' on initialization.

I'm working through a fix though...

stevevanhooser:
(this fix is on the DID side)

stevevanhooser:
OK, the fix is applied on the DID side, not yet tested with NDI

ehennestad:
Looks good as far as I can see.

stevevanhooser:
After the merge, all ndi.test.testsuite.* tests pass for me

PR #293, opened 2024-10-20
changes to make parallel with DID v080

stevevanhooser:
I made a few changes to keep things parallel with DID v080

In DID v080, I changed the `readjsonfilelocation` method in `did.document` so that it no longer accepts partial paths, like `stimulus/stimulus_presentation`. So I had to remove a lot of those. I used to use those partial paths, thinking it would help a programmer find the document, but gave it up, and I think having the system search for so many possibilities might make it slower.

It also requires an update to `vhlab-toolbox-matlab` so that `vlt.file.findfilegroups` can deal with either strings or character arrays. I added an argument block so it is always converted to a char.


stevevanhooser:
The `ndi.test.testsuite.ndi_testsuite` all passes

stevevanhooser:
I was trying to think if there is a reason I should leave this as a request rather than just merging it..I can't think of a good reason so I'll just merge.

PR #294, opened 2024-10-22
small updates to match DID-matlab new changes

stevevanhooser:


stevevanhooser:
vhlab-toolbox-matlab also needs to be pulled (I updated it..interesting that Matlab doesn't allow 'ieee-le' for machine precision with the validator! I always wrote that because I thought it was easier to read (it is clearly a machine precision rather than a 1 or something else)

ehennestad:
> vhlab-toolbox-matlab also needs to be pulled (I updated it..interesting that Matlab doesn't allow 'ieee-le' for machine precision with the validator! I always wrote that because I thought it was easier to read (it is clearly a machine precision rather than a 1 or something else)

It might be this function (which I added) that is the cause of `ieee-le` not being supported (?): 
https://github.com/VH-Lab/DID-matlab/blob/d0ab8ee6667f83dbc9a8d7301fe7de0c7c3e5268/code/%2Bdid/%2Bfile/mustBeValidMachineFormat.m#L1-L14

I think I only did a quick scan in DID to see if it would cover all cases, but that was short-sighted. Will update the function to cover all options

ehennestad:
I added the following commits:
d318e6578cf1fb83afd2278cb07fa8fa39267814 : replace `did.datastructures.assign` with `vlt.data.assign`

8b46fe0589cc343a7c8e51879ef06ca5b86e0e7d : Remove space from byte order

These should fix the issues of this PR without changing the file format/byteorder, which I think should work with PR#83 to DID_MATLAB

I suggest to close this PR without merging

stevevanhooser:
Ah, I should have looked. I thought it was a Matlab function :) Yeah, maybe add all the options described in help fopen ?via mobileOn Oct 22, 2024, at 3:53 AM, ehennestad ***@***.***> wrote:﻿

vhlab-toolbox-matlab also needs to be pulled (I updated it..interesting that Matlab doesn't allow 'ieee-le' for machine precision with the validator! I always wrote that because I thought it was easier to read (it is clearly a machine precision rather than a 1 or something else)

It might be this function (which I added) that is the cause of ieee-le not being supported (?):
https://github.com/VH-Lab/DID-matlab/blob/d0ab8ee6667f83dbc9a8d7301fe7de0c7c3e5268/code/%2Bdid/%2Bfile/mustBeValidMachineFormat.m#L1-L14 [github.com]
I think I only did a quick scan in DID to see if it would cover all cases, but that was short-sighted. Will update the function to cover all options

stevewds:
this is redundant now, not needed

PR #301, opened 2024-11-22
Update ndi.session.dir to allow construction with no input arguments

ehennestad:
Allows the use of array expansion of class objects as described here in the **"No Input Argument Constructor Requirement section"**:

https://se.mathworks.com/help/matlab/matlab_oop/class-constructor-methods.html#btn2kiy


stevevanhooser (approved):
Thanks!

PR #303, opened 2024-11-28
Merge outlines for new epoch class

ehennestad:


+ndi/epoch.m
```diff
@@ -0,0 +1,67 @@
...
+    % 'epoch_clock'             | A cell array of ndi.time.clocktype objects that describe the type of clocks available
+    % 't0_t1'                   | A cell array of ordered pairs [t0 t1] that indicates the start and stop of the epoch for each
+    %                           |   respective epoch_clock{}. The time units of t0_t1{i} match epoch_clock{i}.
+    % 'object'                  | The ndi.epochset object that has the epoch
+    % 'underlying_epochs'       | An array of the ndi.epoch objects that comprise this epochs.
+
+			% note: will need to change underlying_epochs.underlying to underlying_epochs.object in all code
+
+    properties (SetAccess = private)
+        epoch_number (1,1) double {mustBeInteger(epoch_number)} = 0;
```
ehennestad, line 18: No need to specify the property name when the validation function has no other input args, so the following also works.
`epoch_number (1,1) double {mustBeInteger} = 0;`

+ndi/epoch.m
```diff
@@ -0,0 +1,67 @@
...
+        epoch_id (1,1) string = "" % Unique identifier for the epoch; could be any string 
+        epoch_session_id (1,1) string {did.ido.isvalid(epoch_session_id)} = ndi.ido.unique_id % Identifier for the session containing this epoch
```
ehennestad, line 20: Not sure if there should be a default value for this. I.e you will anyway always override that value when creating an object

+ndi/epoch.m
```diff
@@ -0,0 +1,67 @@
...
+        epochprobemap (1,1) {mustBeA(epochprobemap,'ndi.epoch.epochprobemap')} =  ndi.epoch.epochprobemap
```
ehennestad, line 21: You can use any class names for type restriction. I think mustBeA is mostly useful either if the property is not restricted to one specific type, i.e `value (1,1) {mustBeA(value, ["string", "numeric"])}` or for the other problem you had of not defining property of the same type as the class itself
epochprobemap (1,1) ndi.epoch.epochprobemap

+ndi/epoch.m
```diff
@@ -0,0 +1,67 @@
...
+        epoch_clock cell = {} % A cell array of ndi.time.clocktype objects that describe the types of clocks available
```
ehennestad, line 22: Ideally this should be restricted to `ndi.time.clocktype`. If that requires a lot of refactoring, consider keeping it a `cell`

+ndi/epoch.m
```diff
@@ -0,0 +1,67 @@
...
+        t0_t1 cell = {} % A cell array of ordered pairs [t0 t1] that indicates the start and stop time of the epoch in each clock
```
ehennestad, line 23: One idea I had here would be to split this into two properties, i.e `tStart` and `tStop`. But there are a lot of dependencies on t0_t1 so maybe not worth the effort at this time

+ndi/epoch.m
```diff
@@ -0,0 +1,67 @@
...
+        object (1,1) {mustBeA(object,'ndi.ido')} = ndi.ido
```
ehennestad, line 24: Not sure what `object` refers to here. Maybe `parent_epochset_id`?

+ndi/epoch.m
```diff
@@ -0,0 +1,67 @@
...
+        underlying_epochs {} = [] % An array of the ndi.epoch objects that underlie/comprise this epoch
```
ehennestad, line 25: Could to this:       
`underlying_epochs (1,:) {mustBeA(underlying_epochs, {'double', 'ndi.epoch'})} = []
`

or even better to use the custom validation function you added:
`underlying_epochs (1,:) {ndi.epoch.mustBeEpochOrEmpty} = []`

+ndi/epoch.m
```diff
@@ -0,0 +1,67 @@
...
+             %                           |   for each respective epoch_clock{}. The time units of t0_t1{i} match epoch_clock{i}.
+             % 'object'                  | The ndi.epochset object that has the epoch
+             % 'underlying_epochs'       | An array of the ndi.epoch objects that comprise this epochs.
+                  try
+                      for i=1:2:numel(varargin),
+                          obj.set(varargin{i},varargin{i+1});
+                      end
+                  catch
+                      error(['Error in creating ndi.epoch object: ' lasterr ]);
+                  end
```
ehennestad, line 52: arguments
    propValues.?ndi.epoch
end

obj.set(propValues)

+ndi/epoch.m
```diff
@@ -0,0 +1,67 @@
...
+    end % methods
+
+    methods(Static) 
+        function mustBeEpochOrEmpty(value)
+            % mustBeEpochOrEmpty - validate that a value is either an ndi.epoch or is empty
+            %
+            % Syntax:
+            %   ndi.epoch.mustBeEpochOrEmpty(value)
+            % Inputs: <value>, an input
+                assert( isa(value,'ndi.epoch') | isempty(value), 'Value must be an ndi.epoch or be empty.');
```
ehennestad, line 64: Would use `||` instead of `|` since both of the expressions evaluate to a scalar logical:
`assert( isa(value,'ndi.epoch') || isempty(value),`

+ndi/epoch.m
ehennestad: `epochprobemap` can be empty, but can it also have more than one element?
should size restriction be set to `(:,1)`

+ndi/epoch.m
stevevanhooser: I like that idea in general but maybe too much work for too little reward right now.

+ndi/epoch.m
stevevanhooser: Bummer..I guess this doesn't work because `matlab.mixin.SetGet` is not the same class?

But I guess I mean `SetAccess=protected`. That should be sufficient. Maybe that works? I'll try it out.

+ndi/epoch.m
stevevanhooser: OK

+ndi/epoch.m
stevevanhooser: Sounds good

+ndi/epoch.m
stevevanhooser: This allows calling the function with no arguments to succeed. What do you think? Do we need that?

+ndi/epoch.m
stevevanhooser: On the first point, there are some descendent classes of `ndi.epoch.epochprobemap` that might be passed. I was trying to write something that would validate in that case. I don't know if we always want to cast it to `ndi.epoch.epochprobemap` (which might be okay).

On the second point, yes, I mean a vector!

Thanks

+ndi/epoch.m
stevevanhooser: I never know what to do here. I am envisioning the possibility that there may be future descendent classes of `ndi.time.clocktype`, and then it would be impossible to put them in an array. Any ideas?

+ndi/epoch.m
stevevanhooser: I was imagining that it would be the `epochset_object` itself. For underlying epochs, we have passed the underlying object, so that programs can continue down to the root source of the epoch. But I don't like the term `object`, I like `'epochset_object` better, more descriptive, as you point out.

stevevanhooser: It also should be an `ndi.epoch.epochset` object. I had used `ndi.ido` because I didn't want to go through the trouble of making an empty `ndi.epoch.epochset` object but it works if called with no input arguments.

+ndi/epoch.m
stevevanhooser: Yes, I made it as you say except one column instead of one row

+ndi/epoch.m
stevevanhooser: It works with `protected`

+ndi/epoch.m
stevevanhooser: I now included code that makes sure that if any properties are specified in the creator, then they all are

+ndi/epoch.m
stevevanhooser: Actually, it works for me with `SetAccess=private`. I accidentally ran the old version and it worked.

+ndi/epoch.m
ehennestad: `matlab.mixin.Heterogeneous`

stevevanhooser: Wow that is magic

PR #306, opened 2024-12-16
Update element - Add Heterogeneous mixin and support 0-argument creation

ehennestad:


stevevanhooser:
Thanks, all tests pass after merging as well

PR #310, opened 2025-01-08
Replace try block with if/else ndi.file.navigator/epochid

ehennestad:
Should do the same, but explicitly checks if file is ingested instead of using a `try` statement that is more expensive, especially if an error is caught, i.e when files are not ingested


ehennestad:
(Added as PR in case I am overlooking something)

stevevanhooser (approved):
looks good..I wonder why I did it that way...it kind of looks like I did it for a reason but maybe it was because I hadn't written isingested yet? Let's go with it

PR #317, opened 2025-03-13
Remove the status output for `ndi.cloud.api*` functions

ehennestad:
Remove the status output for `ndi.cloud.api*` functions because it will always be 0. If any of these functions fail, an error will be thrown, and status will not be returned.

Most changes are related to the title, but some bugs (mostly related to renamed namespaces that has not been updated) were also fixed

All tests are passing


stevevanhooser (approved):
Thanks!

PR #325, opened 2025-04-15
Ensure global path constants are updated

ehennestad:
Ensures global path constants are updated in DID before calling DID's `did.document.readjsonfilelocation` method which relies on some NDI path contants being available in the DID `PathConstant` object.


stevevanhooser (approved):
Thank you!

PR #326, opened 2025-04-16
Downsample update

jesshaley:


+ndi/+element/downsample.m
```diff
@@ -73,4 +85,5 @@
     % Add the downsampled data to the new object
     elem_out.addepoch(ndi_element_timeseries_obj_in.epochid(epoch_i),...
         epoch_clock_str,t0_t1, t_down, data_down);
-end
+    D.cache.clear(); clearAllMemoizedCaches;
```
stevevanhooser, line 88: Let's not clearAllMemoizedCaches here (D.cache.clear() does it, and maybe some day we'll be able to figure out how to avoid clearing them all)

stevevanhooser:
Thanks @jesshaley I made a couple comments

jesshaley:
`ndi.element.downsample` and `ndi.element.oneepoch` have been updated to check for new files in the session and only downsample or append those that do not yet exist.  `ndi.element.oneepoch` is also now able to handle the addition of a missing epoch in the middle of the session by checking the timestamps and sorting the data accordingly. 

Note: `mlt.spectrogram` in the NDIcalc-marder-matlab repo does not yet consider non-consecutive and/or variable sample rate time stamps.

stevevanhooser:
The spectrogram calculation can just erase and replace. (The epochs should be added in alphanumeric order so the big wholething epoch should always be in order, right?)

jesshaley:
> The spectrogram calculation can just erase and replace. (The epochs should be added in alphanumeric order so the big wholething epoch should always be in order, right?)

What happens in the event that:
1. An epoch file is misplaced and only later added? (e.g. a 10 min recording file was not in the directory when the epoch table is originally made, but is later added).
2. There is a gap in the recording? (e.g. X mins of recording do not exist) Or, similarly, the sampling rate is not consistent?

Another note: Curiously, the new `ndi.element.oneepoch` function creates a slightly longer time series. Specifically, for the session directory I worked on there were 785 additional time points at the end of the file. The first 1400508 time points are identical.

stevevanhooser:
Hmm, that's interesting. I wonder where those extra 785 points are coming from. Are those points really there in the last epoch (if you look at it individually)?

> ndi.element.oneepoch is also now able to handle the addition of a missing epoch in the middle of the session by checking the timestamps and sorting the data accordingly

This is cool, but what happens if the epoch only has dev_local_time ? Then we can't know the overall time in the previous oneepoch, right? I don't think it would be able to handle an epoch dropped in the middle the way that it can when there is also an exp_global_time clock available. The function should work in all cases. I guess we can detect that it falls short if the total time of the oneepoch does not equal the sum of all the individual epochs, right? 

It would also be a nice feature to add an option (name/value option) to force it to startOver (=false by default, = true if the user sets it) so that the user can force it to start over (discard what it has). If the user knows they've messed up the epoch order and they don't have any global clock, for example.

jesshaley:
I ran it again and was not able to reproduce the mismatch. Both the new and old versions of oneepoch now show those 785 points which do exist in the last downsampled epoch.

As for the clocks, that makes sense! Do you think the following strategy work with the additional clock types? If so, which clock types would work for defining time stamps vs. appending?
```
if one of the clocks is exp_global_time, utc, or dev_global_time
   Use that clock time to define time stamps and sort accordingly
elseif one of the clocks is dev_local_time
   assume epochs are in order and append; allow user to specify startOver
else
   error('no clock to keep track of epochs')
```

Clock types:
```
% 'utc'                     | Universal coordinated time (within 0.1ms)
% 'approx_utc'              | Universal coordinated time (within 5 seconds)
% 'exp_global_time'         | Experiment global time (within 0.1ms)
% 'approx_exp_global_time'  | Experiment global time (within 5s)
% 'dev_global_time'         | A device keeps its own global time (within 0.1ms)
%                           |   (that is, it knows its own clock across recording epochs)
% 'approx_dev_global_time'  |  A device keeps its own global time (within 5 s)
%                           |   (that is, it knows its own clock across recording epochs)
% 'dev_local_time'          | A device keeps its own local time only within epochs
% 'no_time'                 | No timing information
% 'inherited'               | The timing information is inherited from another device.
```

jesshaley:
@stevevanhooser I still need to create a unittest, but `oneepoch` is close to being finished. However, I've received the following error while trying to search and subsequently remove `oneepoch` documents from the database. Any ideas on how to fix this? Maybe something special about the `element_epoch` document class that I need to copy?

Unrecognized field name "ndi_oneepoch_class".

Error in getfield (line 14)
    f = s.(field);

Error in ndi.database.fun.ndi_document2ndi_object (line 40)
        obj_string = getfield(obj_struct,['ndi_'
        obj_parent_string '_class']);

Error in ndi.session/getelements (line 711)
                elements{i} =
                ndi.database.fun.ndi_document2ndi_object(doc{i},
                ndi_session_obj);

Error in downsampleTest (line 19)
e = S.getelements();

stevevanhooser:
Ah, yes..add a field to that class called ’ndi_oneepoch_class’ and give it the value ’ndi.element’ . It is trying to build a Matlab object from the document and it expects them to have a field with that name and an instruction as to the object class it should try to create.

jesshaley:
I tried adding that, but am still having problems retrieving files using `S.getelements`. Any ideas how to remedy? 

Error using ndi.element (line 88)
This document does not have parameters 'element'.

Error in ndi.database.fun.ndi_document2ndi_object (line 43)
    o = eval([obj_string '(ndi_session_obj,
    ndi_document_obj);']);

Error in ndi.session/getelements (line 711)
                elements{i} =
                ndi.database.fun.ndi_document2ndi_object(doc{i},
                ndi_session_obj);

Error in downsampleTest (line 19)
e = S.getelements();

+ndi/+database/+fun/finddocs.m
```diff
@@ -0,0 +1,35 @@
+function [docs] = finddocs(S,elementID,epochID,documentType)
```
stevevanhooser, line 1: Please rename this function to be something like findDocTypeElementIDEpochID because this finds very specific kinds of documents. (I would imagine something called findDocs to have a more general function.) 
Thanks!

stevevanhooser (approved):
Looks good, thanks!

PR #330, opened 2025-04-22
Updated epochclocktimes definition and schema

jesshaley:


ndi_common/database_documents/epochclocktimes.json
```diff
@@ -10,14 +10,9 @@
 			{ "definition":					"$NDIDOCUMENTPATH\/epochid.json"}
 		]
 	},
-	"epochclocktimes": {
```
stevevanhooser, line 13: So, this was actually a deliberate double step. It allows epochclocktimes to contain an array of values epochclocktimes.

Our root property lists cannot be arrays, so we need the nested list.

jesshaley: I see. When I was using `epochclocktimes` for the beats document, I only saved one clocktype matching the data saved to the `beats.vhsb`. For example, I have `exp_global_time`, so onset and offset times were saved as 'datenum' values and `epochclocktimes = struct('clocktype','exp_global_time','t0_t1',[t0,t1])`.

However, would it be better convention for me to convert the beats onset and offset times to the dev_local_time? If so, is epochclocktimes even necessary for a beats doc given that the beats doc is associated with an element_id and epoch_id which have the associated clocktypes and t0_t1?

jesshaley:
Okay! Just to double check, for the data being saved to beats.vhsb, should I have it always (and only) use the dev_local_time clock?

stevevanhooser:
Yep

jesshaley:
This is the general architecture for the ppg_beats pipeline as I'm currently thinking about it:

`mlt.wholeDayHeartBeatsDoc `takes a session object, name_out, ref_out that references a single epoch element.
1. Read `timeseries` (use `dev_local_time` regardless of availability of global clock) and `zscore`.
2. Get `beats` data using `mlt.detectHeartBeatsImproved()`
3. Add `beats` document to database.

`mlt.graceHeartBeatPlotDoc` takes a session and loops through all probes
1. Read timeseries (use global if available, if not local)
2. Load `beats` data from document and convert from local to global (if necessary).
3. Plot time, data, and beats using `mlt.gracePlotHeartBeat()`.

Remaining questions:
- Does this match your thinking?
- Is attaching `epochclocktimes` to beats necessary if the `beats` doc necessitates association with an `element`.timeseries which contains that info?

jesshaley:
This branch should be ready to merge.

stevevanhooser (approved):
Thanks!

PR #337, opened 2025-04-28
Adding unit test for oneepoch

jesshaley:
@stevevanhooser having some problems with ingestion of white matter data...
1. Bugs with reading sample rate in `+ndi/+daq/+reader/+mfdaq/ndr.m` and `+ndi/+daq/+reader/mfdaq.m` (see quick fixes in changed files on this branch).
2. Bugs with handling `int16` data. Try running the `OneEpochTest`. It breaks on `S.ingest()` due to issues in `+ndi/+daq/+reader/mfdaq.m` starting with `underlying_datatype()`.


stevevanhooser:
Thanks..there are some changes on main that this branch doesn’t have to ndi.daq.reader.mfdaq.ndr.. If you update it will probably fix the issue
Thanks!
Steve

------------------------------------------
Stephen D. Van Hooser, PhD
Associate Professor of Biology
Brandeis University
(+1) 781-736-2721
{www <http://www.vhlab.org/>,wiki <http://wiki.vhlab.org/>,code <http://code.vhlab.org/>,dataclass <http://dataclass.vhlab.org/>,twitter <http://twitter.vhlab.org/>}.vhlab.org <http://vhlab.org/>

jesshaley:
Unless I'm doing something wrong, the int16 issue seems to still exist.

```
Error using  .* 
Integers can only be combined with integers of the same class, or scalar doubles.

Error in ndi.compress.scaled2underlying

Error in ndi.daq.reader.mfdaq/ingest_epochfiles (line 793)
                                data = ndi.compress.scaled2underlying(data, mypoly);
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in ndi.daq.system/ingest (line 431)
                    new_d = ndi_daqsystem_obj.daqreader.ingest_epochfiles(ef,et(i).epoch_id);
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in ndi.session/ingest (line 459)
                [b_here,daq_d{i}] = daqs{i}.ingest();
```

`mypoly` is a double while `data` is int16. 
https://github.com/VH-Lab/NDI-matlab/blob/59568d2bb2128bef4e3058ffdc2380a05b4c3de6/%2Bndi/%2Bdaq/%2Breader/mfdaq.m#L752

jesshaley:
@stevevanhooser I'm still wrapping this up, but the unittest revealed an interesting bug. When the data is created in `ndi.element.oneepoch`, the class matches the input elements. However, after the oneepoch data is added to the element the class is converted to double. Am I doing something wrong? Or is this a bug in `ndi.element.timeseries`?

```
% class(data) = 'int16'
elem_out = ndi.element.timeseries(D, name_out, reference_out,...
        ndi_element_timeseries_obj_in.type,ndi_element_timeseries_obj_in,false);
[elem_out,epochdoc] = elem_out.addepoch(epoch_id,...
    strjoin(ecs,','), t0_t1, time, data, strjoin(epoch_ids,','));
[data,time] = elem_out.readtimeseries(1,-inf,inf);
% class(data) = 'double'
```

jesshaley:
@stevevanhooser I've completed this draft of a OneEpochTest. I'm not sure if I've maintained all the best practices (it seems difficult to find literature on this), but I used fixtures that will be reusable and included some parameterization of test parameters. The unittest still [fails while checking the class of the oneepoch elements as mentioned in my previous comment](https://github.com/VH-Lab/NDI-matlab/pull/337#issuecomment-2840424460).

tests/+ndi/+unittest/+element/OneEpochTest.m
```diff
@@ -0,0 +1,183 @@
...
+
+                % Get data
+                probe_data = [];
+                for j = 1:numel(probe_et)
+                    probe_data = cat(1,probe_data,probe.readtimeseries(j,-Inf,Inf));
+                end
+                oneepoch_data = oneepoch.readtimeseries(1,-Inf,Inf);
+                testCase.verifySize(oneepoch_data,size(probe_data),...
+                    'OneEpoch does not return the expected number of samples.')
+                testCase.verifyClass(oneepoch_data,class(probe_data),...
```
stevevanhooser, line 176: This is failing for me, too, for unknown reasons. When I went into the debugger, oneepoch_data is in fact a double. So somehow the command isn't working, I don't quite understand. 

But I think a better test might be:

            testCase.assertEqual(a, b_expected, ...
                'Values should be equal, treating NaNs as equal.', ...
                'TreatNaNsAsEqual', true);

This not only tests the class and the size but also whether the values are equal (allowing NaNs to be equal as well; by default any NaN makes equal false). 

Does that make sense?

jesshaley:
That makes sense, though may fail if global time is used and epochs are
non-continuous.

However, shouldn’t oneepoch_data be int16 since white matter data is int16
(and thus probe_data is int16)?

stevevanhooser:
Oh I see. oneepoch_data will be double because I think it is stored by addepoch as double when it is written to the binary file:

https://github.com/VH-Lab/NDI-matlab/blob/59568d2bb2128bef4e3058ffdc2380a05b4c3de6/%2Bndi/%2Belement/timeseries.m#L115C13-L115C52

vhsb can use other datatypes but we haven't engaged them (we'd have to pass those parameters with name/value pairs).
https://github.com/VH-Lab/vhlab-toolbox-matlab/blob/2fc776b388e9881f6eee02cdf407a09a2ae80ff2/%2Bvlt/%2Bfile/%2Bcustom_file_formats/vhsb_write.m#L20

So I guess it won't be equal, although you could cast the directly-read probe data to double and then test, since you know element_epochs will be double.

Instead of testing the entire epoch, you could pick a few values that are easy to calculate. If you can easily calculate the value of samples 1500, 1550, 2500, 10000, for example, then you could only test those values, and you might not have to rebuild the whole one epoch vector again from the probe epochs (with the fill-in).

jesshaley:
@stevevanhooser okay! That makes sense. I'll update that code.

stevevanhooser:
Is this ready to be merged?

jesshaley:
Almost. I need to fix the verifyClass issue. I’ll have this done today and
will send you a notification.

jesshaley:
@stevevanhooser this should be ready to merge now.

stevevanhooser (approved):
Thanks! Works on my end

PR #342, opened 2025-05-04
Dabrowska session maker

jesshaley:
`obj = sessionMaker(path,variableTable,options)` and `[sessions,ind] = sessionIndices(obj)` seem to be working. `addDaqSystem(obj,daqName,reader,navigator)` does not yet work. I am still working on this and need to add/edit documentation to several of the functions that I edited. However, I think it would be helpful to check-in about some of the following:

The `variableTable` produced by `ndi.setup.conv.datalocation.processFileManifest` does not maintain the full path. I added an input to `sessionMaker()` to circumvent this, but would it be better to have `processFileManifest` return the full path somewhere in the table?

Is there a strong reasoning behind returning `NaN` values for invalid `SessionRefs`? This creates extra work when processing unique values of `variableTable.SessionRef`. I’ve done this and it works fine, but maybe it would be better to leave these rows as empty char arrays and have an additional logical variable in the table that denotes valid sessions?

It seems a little strange to me that `variableTable.IsExpMatFile` is a cell array rather than just a logical denoting if the file extension is .mat.

I could not figure out a way to delete the session to create a new `sessionRef` (if needing to overwrite). I added `erase` functions to `ndi.database` and `ndi.session` that delete the entire .ndi folder allowing this to occur. Please let me know if there is a better approach.

I ran into an issue with `mksqlite` at ~19 sessions in. I got the error:
```
Error using mksqlite
no free database handle available
```
My current work around is to run `mksqlite('close')` after the creation of each session. This seems to solve the problem, but let me know if there is a better way.

From my understanding, a `navigator = ndi.file.navigator` requires association with a particular session. How do we want to therefore pass the navigator to the function addDaqSystem?

### DIFF
--- a/+ndi/+dataset/dir.m
+++ b/+ndi/+dataset/dir.m
@@ -51,4 +51,29 @@ classdef dir < ndi.dataset
             end;
         end; % dir(), creator
     end; % methods
+
+    methods (Static)
+        function dataset_erase(ndi_dataset_dir_obj, areyousure)
+            % DATABASE_ERASE - deletes the entire session database folder
+            %
+            % DATABASE_ERASE(NDI_DATASET_DIR_OBJ, AREYOUSURE)
+            %
+            %   Deletes the session in the database.
+            %
+            % Use with care. If AREYOUSURE is 'yes' then the
+            % function will proceed. Otherwise, it will not.
+            arguments
+                ndi_dataset_dir_obj {mustBeA(ndi_dataset_dir_obj,'ndi.session.dir')}
+                areyousure (1,:) char = 'no';
+            end
+
+            if strcmpi(areyousure,'yes')
+                rmdir(fullfile(ndi_dataset_dir_obj.path,'.ndi'),'s'); % remove database folder
+            else
+                disp('Not erasing session directory folder because user did not indicate they sure.');
+            end
+            delete(ndi_dataset_dir_obj);
+
+        end % dataset_erase()
+    end % methods (Static)
 end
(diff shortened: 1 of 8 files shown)


stevevanhooser:
Lots of great questions here:


> The variableTable produced by ndi.setup.conv.datalocation.processFileManifest does not maintain the full path. I added an input to sessionMaker() to circumvent this, but would it be better to have processFileManifest return the full path somewhere in the table?

I think we will need the second input, possibly an optional input. The reason is that sometimes the file manifest may come from a different computer's view, in which case the prefix might be all messed up. For example, the Dabrowska files on my computer are at /Users/vanhoosr/Library/CloudStorage/GoogleDrive-steve@walthamdatascience.com/Shared drives/Dabrowska but on yours are probably at /Users/vanhoosr/Library/CloudStorage/GoogleDrive-jess@walthamdatascience.com/Shared drives/Dabrowska. I think allowing for different prefixes is helpful, if a little burdensome sometimes.

> Is there a strong reasoning behind returning NaN values for invalid SessionRefs? This creates extra work when processing unique values of variableTable.SessionRef. I’ve done this and it works fine, but maybe it would be better to leave these rows as empty char arrays and have an additional logical variable in the table that denotes valid sessions?

These variables come right from our regular expression matcher. There isn't any computational intervention afterwards. So there is either a match or there isn't. For some files, there isn't a corresponding session (like notes and things). We want our sessionMaker to be robust to that, because we are going to be vacuuming up file lists that contain all sorts of things.

> It seems a little strange to me that variableTable.IsExpMatFile is a cell array rather than just a logical denoting if the file extension is .mat.

Same, it's just a regular expression matcher output.

jesshaley:
sessionMaker almost completed. A [DID bug](https://github.com/VH-Lab/DID-matlab/issues/100) needs to be fixed for addDaqSystem to work.

stevevanhooser (approved):
This looks good at this point..we will do a more thorough review in the dabrowska branch

PR #345, opened 2025-05-07
Dabrowska

jesshaley:
@stevevanhooser epochProbeMapMaker now working


stevevanhooser:
@jesshaley : Is this warning okay/expected? (ProbePostfix)

> Warning: The new variables being added to the table have fewer rows than the table. They have been
> extended with rows containing default values. 
>

jesshaley:
The warning is expected. But in retrospect, I will fix that so that the
warning doesn’t occur. It’s just referring to saving a variable to the
table and leaving the non mat file rows blank (which is what we want).

jesshaley:
@stevevanhooser I added regex expressions to process the optogenetics experiments and was able to combine all experiments into one session (see `working.m`) . The code mostly works, but here are a couple issues I ran into.
1. The behavior of processFileManifest is slightly strange. When no rows match the regex, the column is returned as a double filled with NaN. When at least one row matches the regex, the column is returned as a cell array. I've created a simple work around for now.
2. How do we want to handle the TLS stimulus?

stevevanhooser:
Awesome, looking good here

Two questions and a note when you have a chance @jesshaley:

1. There is a `filename` assignment missing (used at the end of `working.m`)
2. It looks like there might be an alternative copy of the files? You could put that back on Google Drive so I could take a look.
3. I started a synthesizing file `ndi.setup.conv.dabrowska.doImport` that incorporates a lot of what was in `working.m` but a little more formally (it handles paths a little more generally). Shall we transition to this file?

stevevanhooser:
(@jesshaley)

jesshaley:
Yes! Go for it

stevevanhooser (approved):
Thanks all

PR #356, opened 2025-05-15
Fix AND operator for scalar values in ndi.document

ehennestad:
I got this error due to using `&` instead of `&&` in check for whether a document has dependencies.

```matlab
Unrecognized field name "depends_on".

Error in ndi.document/dependency (line 261)
                ~isempty(ndi_document_obj.document_properties.depends_on);
```

### Example:

Does not work:
``` matlab
>> S = struct('a', 1);
hasB = isfield(S, 'b') & ~isempty(S.b)
Unrecognized field name "b".
```

works:
```matlab
>> S = struct('a', 1);
hasB = isfield(S, 'b') && ~isempty(S.b)

hasB =

  logical

   0
```

Made replacement in other places as well, as recommended when comparing scalar values

### DIFF
--- a/+ndi/document.m
+++ b/+ndi/document.m
@@ -257,7 +257,7 @@ classdef document
             %
             names = {};
             depend_struct = vlt.data.emptystruct('name','value');
-            hasdependencies = isfield(ndi_document_obj.document_properties,'depends_on') & ...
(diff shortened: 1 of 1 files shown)


stevevanhooser (approved):
Thanks!

PR #360, opened 2025-05-16
Improve warning reset

ehennestad:
Just a small fix, I ended up having all my warnings turned off because I errored out of the code before the warnings were reset (in `ndi.dataset.dir`)


stevevanhooser (approved):
Thanks!

PR #364, opened 2025-05-19
Creating blank from schema

msalvi1:


+ndi/+document/buildunifieddoc.m
```diff
@@ -0,0 +1,103 @@
...
+%NDI_BUILDUNIFIEDDOC Builds a blank NDI document from a schema file and saves it as a JSON file,
+%   optionally embedding schema information.
+%
+%   GENERATED_JSON_PATH = NDI_BUILDUNIFIEDDOC(SCHEMA_FILE_PATH, OUTPUT_DIR, EMBED_SCHEMA_INFO)
+%   takes the file path to an NDI schema JSON file, an output directory, and a boolean flag
+%   indicating whether to embed schema information into the blank document.
+%
+%   Input:
+%       SCHEMA_FILE_PATH (char or string): Path to the NDI schema JSON file.
+%       OUTPUT_DIR (char or string): Path to the directory where the JSON
```
stevevanhooser, line 11: This function should not generate a file, but instead should generate a Matlab structure that sits at 

d.document_properties

where d is an ndi.document object.

To see what one looks like, try

d = ndi.document('base')

One should be able to call

d = ndi.document(output_structure)

where output_structure is the output structure from buildunifieddoc.

(see `help ndi.document` to read about the constructor)

+ndi/+document/buildunifieddoc.m
```diff
@@ -0,0 +1,103 @@
...
+        % --- Embed depends_on information and create blank values ---
+        unified_doc.depends_on = [];
+        if isfield(schema, 'depends_on') && ~isempty(schema.depends_on)
+            unified_doc.depends_on = schema.depends_on; % Embed the constraints
+            for i = 1:numel(unified_doc.depends_on)
+                unified_doc.depends_on(i).value = ''; % Add a 'value' field for the blank value
+            end
+        end
+
+        % --- Create the main property list field with default values and optional schema info ---
```
stevevanhooser, line 65: This code seems not to visit the superclasses to create the superclass properties. I would guess that if you tried to build any other document than 'base' it would not have the fields of other classes (such as base).

+ndi/+document/buildunifieddoc.m
```diff
@@ -0,0 +1,103 @@
...
+%                                    file should be saved.
+%       EMBED_SCHEMA_INFO (logical): If true, embeds some schema information
+%                                    into the blank document.
+%
+%   Output:
+%       GENERATED_JSON_PATH (char or string): Path to the generated JSON file.
+
+    try
+        % Read the schema JSON file
+        schema_text = fileread(schema_file_path);
```
stevevanhooser, line 21: Let's keep this for now, where the function takes as input the schema file directly. Later, we will refactor so that takes the document type (e.g., 'base') and then searches the known document paths to find the json. But let's not do this right now.

stevevanhooser:
Hi @msalvi1 , thanks for the code. I added some comments. Let me know when you'd like me to take a look again!

Best
Steve

msalvi1 (commented):
Implemented the changes from Steve's review. I only tested on three types of schemas: element_schema, subject_schema, treatment_schema but it seemed to correctly create the superclass properties and generate the correct struct for the ndi.document constructor.

PR #367, opened 2025-05-19
added PATO ontology

stevevanhooser:


stevevanhooser:
@jesshaley you can merge when you're at a good spot

PR #370, opened 2025-05-20
Add functions for registering DOIs for cloud datasets

ehennestad:


+ndi/+cloud/+admin/+crossref/+conversion/convertFunding.m
```diff
@@ -0,0 +1,15 @@
+function fundingObj = convertFunding(cloudDataset)
+    
+    if ~isfield(cloudDataset, 'funding') || isempty(cloudDataset.funding)
+        fundingObj = crossref.model.FrProgram.empty;
+    else
+        fundingDetails = cloudDataset.funding;
+
+        frAssertion = crossref.model.FrAssertion(...
+            "Name", "funder_name", ...
+            "Value", string(fundingDetails.source));
```
stevevanhooser, line 10: I got an error here because Jess had multiple funding sources. For now I just took the first one in the list.

ehennestad: Pushed a commit to fix that

stevevanhooser:
Hi @ehennestad 

I posted a couple of errors on our slack thread. It's close!

Thanks
Steve

stevevanhooser:
I was able to successfully submit a DOI this morning just by removing the `language="eng"` part from the line `<database_metadata language="eng">`. So probably just removing that is enough!

ehennestad:
Does that mean you kept the `ai:program` name without changing to `ali:program`?

I am wondering if language="en" would work. There is an example file with language="en". The "eng" i got from the documentation...

stevevanhooser:
Yes, that's right: in the most recent submission I kept it as `ai:program` and it was fine.

Not sure if there is an alternative lang that would pass. Or if it needs to be xml.lang or something like that...

stevevanhooser (approved):
Thanks! We already tested so I am going to pull it in (if there are any small remaining issues we can fix later)

PR #374, opened 2025-05-28
Update `ndi.cloud.authenticate`

ehennestad:
- Added default value for input argument to local function
- Added mode to disable interactive steps


stevevanhooser (approved):
Thanks!

PR #375, opened 2025-05-29
Created ontologySet database and schema docs

jesshaley:


ndi_common/schema_documents/data/ontologySet_schema.json
```diff
@@ -0,0 +1,42 @@
...
+			"queryable": 1,
+			"documentation": "The shortened name(s) of the data variable(s) in a comma-seperated list. Must be valid MATLAB variable names."
+		},
+		{
+			"name": "ontologyNodes",
+			"type": "string",
+			"default_value": "",
+			"parameters": "",
+			"queryable": 1,
+			"documentation": "The ontology node id(s) of the data variable (s) in a comma-seperated list."
```
stevevanhooser, line 29: Maybe here it would be a good idea in the documentation to indicate the form (sometimes ontologies use a colon-based approach others use an underscore, but we will use ontology:nodeID)

jesshaley:
@stevevanhooser added detailed examples for each field. Let me know if we need any further changes.

jesshaley:
@stevevanhooser ready for review. EPM and FPS ontologyRowDocuments are created. The new ontology terms (for FPS and other general identifiers) need to be significantly audited, but the code runs with the placeholders I/gemini created.

You can try running it using the code in `+ndi/+setup/+conv/+dabrowska/working.m`.

Note: there is a bug when using `ndi.query` in `tableDocMaker.createOntologyTableRowDoc` to find existing documents using unique identifiers. I can't figure out what is wrong.

stevevanhooser:
Looking good!

1) I modified the ontologies a little bit and made a new term
elevatedPlusMazeTestDuration
This would store the test duration (300 seconds in this case)
and modified the ones that are % time in 5 minutes to be % time in test duration
I modified these only in the ontology; could you please update in your tables?

2) Can you please make a subject table for these animals? We can then put it through subjectMaker to make the subjects.

stevevanhooser:
check out the new function to find all the variables in the dataset / session!

+ndi/+gui/+component/ProgressBarWindow.m
```diff
@@ -394,14 +409,14 @@
                 app.ProgressGrid.RowHeight);
             app = app.setFigureSize(sum(rowHeight));
 
-            % Throw error if terminated in the middle of task
+            % Throw warning if terminated in the middle of task
```
jesshaley, line 412: @stevevanhooser I had made these errors instead of warnings based on our discussion about wanting the X button to cause termination of the code. If it's a warning, it doesn't actually do anything. What behavior do we want?

stevevanhooser: OK it can be turned back. In some of my testing a time-out threw an error after I had already started another job and it stopped that job. So maybe the time-out should not throw an error but manual closing can / should.

Thanks!

jesshaley:
@stevevanhooser this should be ready to review

stevevanhooser (approved):
Looks good, thanks!

PR #376, opened 2025-05-30
Editing the obo in cline, test driving cline really

stevevanhooser:


andreagaede:
I think this empty.obo file has to be integrated into the one in the empty ontology scaffold, because that file is set up to dynamically import from other ontologies. Right now it is just set up to import from three that the ODK suggested, but we can change it later in Protege I think. The format of the terms is slightly different too. With regards to other parts of the scaffold, I don't think anything has to be changed. I think by using Protege, it makes the other versions of all the necessary files though -- like there is a simple and full obo file that go along with the empty.obo

andreagaede:
@stevevanhooser I added a few terms into my empty ontology. If I add the rest in, will you be able to use that?

andreagaede:
I made a new scaffold called empty and I matched the initial IDs I used to the ones you made up, though I could change them to just count up from 1.

________________________________

PR #383, opened 2025-06-09
App tests2unittests

jesshaley:
1. Added unittest for ndi.app.markgarbage
2. Tried to add unittest for ndi.app.spikeextractor. Function appears to be deprecated


stevevanhooser (approved):
Thanks!

PR #384, opened 2025-06-09
Database tests2unittests

jesshaley:
Replaces all the files in the ndi.test.database folder. One unittest fails (rightfully so) as it points out that contrasttuning_calc.json requires vision/contrast_tuning.json which is missing.


stevevanhooser:
OK, it is a bug that it is in that repo, but it is a bug in that repo. Thanks!

PR #396, opened 2025-07-08
slightly updated messages in upload/download

stevevanhooser:


stevevanhooser:
(these are very minor, just merging them)

PR #398, opened 2025-07-09
added space before tables in docs

stevevanhooser:
This is what I had to do to get the tables to show up in the docs page (add an empty line before the tables)


stevevanhooser:
(just showing you so you see it, it's not a heavy change)

PR #405, opened 2025-07-16
deprecating ndicloud_ontology_lookup (use ndi.ontology.lookup now)

stevevanhooser:
We should be using ndi.ontology.lookup


stevevanhooser:
After either @ehennestad or @jesshaley review this simple change please merge it

+ndi/ontology.m
jesshaley, line 1: @stevevanhooser I thought we wanted all shortnames to be based on the concatenation of the full term name. Are we going to override this default for EMPTY terms?

jesshaley (changes requested):
@stevevanhooser see note about ontology shortnames. Note sure what we want to do.

+ndi/ontology.m
stevevanhooser: It seems like I don't have your latest changes. Do you have changes to this on another branch? The goal is to completely switch to making short names ourselves, yes, using your helper function.

jesshaley: @stevevanhooser I believe this file has my changes. I just added a line under the lookup function that creates the shortname.

stevevanhooser: OK, meaning that `main` already has the latest? And I should just drop the changes here?

jesshaley (approved):
Looks good

stevevanhooser:
oh I'm dumb nevermind. yes you did. (Or I had a stale window??) Thank you!

jesshaley:
@stevevanhooser no problem! I do think I waited a bit while I investigated the failing tests.

PR #408, opened 2025-07-16
adding empty to dabrowska approach terms

stevevanhooser:
I needed this for things to work after switching to ndi.ontology.lookup


stevevanhooser:
Glad you reviewed it, I think maybe I was doing some testing or something

PR #412, opened 2025-07-16
Editing errors in database documents

jesshaley:
@stevevanhooser there were 3 errors in the json document unittests. This branch fixes two of them. The remaining issue is that `NDIcalc-marder-matlab/ndi_common/database_documents/calc/contrasttuning_calc.json` references `$NDICALCDOCUMENTPATH\/vision\/contrast_tuning.json` which is not on the path. I wasn't sure how you might want to resolve this.


stevevanhooser:
I just updated NDIcalc-marder-matlab

stevevanhooser:
No we shouldn't change it. My question was whether or not this is affecting the GitHub action pipeline (which presumably does not by default include NDIcalc-marder-matlab as an installation) or just the run on your computer. (EIther way, we should fix NDIcalc-marder-matlab (I did).)

So I guess my question is really, are we installing NDIcalc-marder-matlab on our GitHub action pipeline? I think we should not have it there.

jesshaley:
@stevevanhooser Ah! You're correct. NDIcalc-marder-matlab is not on the Github action pipeline.

PR #415, opened 2025-07-16
Add `Visible` prop to `ProgressBarWindow` and run progressbar unittests with `Visible = "off"`

ehennestad:
Goal:
- Prevent figure popup when running unit tests locally.
- Skip calling figure on existing figure handle when running unit test on GitHub runner as this causes errors on older MATLAB releases.

All tests are passing.


jesshaley (approved):
Thanks @ehennestad !

PR #416, opened 2025-07-17
Progressbar backwards compatibility

ehennestad:
Add fallback for setting `Icon` property of `uibutton` in the `ProgressBarWindow` class

Using 'error' and 'success' as value for the `Icon` property was introduced in MATLAB R2022b.

At the moment, fallback is no icon. We could merge as is, or spend some time finding suitable icons, @stevevanhooser @jesshaley ?


stevevanhooser:
Maybe just leave it as no icon for now? It would be better to find one but we've got our hands full with so much right now...

PR #420, opened 2025-07-18
Image and video parameter documents

jesshaley:
Creating new documents to hold the metadata for images and videos


jesshaley:
Accidentally added haley-data changes to haley-newdoctypes. Merging now to avoid conflicts.

PR #421, opened 2025-07-21
side-stepping syncgraph when we are asking for the same referent's epoch

stevevanhooser:
This speeds up using syncgraph massively when there is no need to go for the graph. For example, when asking for a single epoch from a probe or element when you already know the element. 

Syncgraph is useful for synchronizing devices across clocks. But we have been loading the whole graph even if we are just looking for an epoch on a single element or device. 

This now bypasses that. For the Marder lab datasets (with tens of thousands of probe-epochs) it was taking quite a long time just to compute the syncgraph unnecessarily.


stevevanhooser:
(Quite a long time == 24 hours)

stevevanhooser:
(Down the road we should switch to a file-based syncgraph rather than a memory-based one)

jesshaley (approved):
Looks good!

PR #429, opened 2025-08-19
Update test-cloud-api.yml

ehennestad:
- [x] Add tests for uploading documents containing inf and nan values
- [x] Add workflow step for publishing annotated test results


stevevanhooser:
@ehennestad Is this ready to be merged do you think?

PR #438, opened 2025-09-01
Updated syntax for compatibility

jesshaley:
Sorry about all the issues Eivind. Gemini lied...hopefully this fixes it?


ehennestad:
> Sorry about all the issues Eivind. Gemini lied...hopefully this fixes it?

No worries! The problem was actually that creation of empty `NaT` using the following syntax was not supported in R2021b:
```matlab
emptyNat = NaT([])
```

So what happened is that the `ndi.fun.table.vstack` silently created an empty `double` instead of an empty `NaT` and the verifications failed due to that.
I fixed that here https://github.com/VH-Lab/NDI-matlab/pull/438/commits/924bae17625d280425df273adf8372c8bb043b2d

 Maybe consider showing a warning here if that fallback is triggered:

https://github.com/VH-Lab/NDI-matlab/blob/ca9ccbba6f61cfeab7fc7c3a60ecec90718f85cd/src/ndi/%2Bndi/%2Bfun/%2Btable/vstack.m#L122-L125

stevevanhooser:
Q: is this ready to merge?

PR #442, opened 2025-09-03
Updates to Haley import

jesshaley:
1. New subjectMaker
2. Raw image and video files added


jesshaley:
@stevevanhooser this is ready for review! I will be sending you the link to the updated zipped dataset shortly. The only remaining task for this will be to check the tutorial once the dataset is online!

PR #621, opened 2026-01-08
Hunsberger import

jesshaley:


jesshaley:
@stevevanhooser there will likely be a few more small changes, but this is probably a good time to merge. Please note several changes to the docTables as well as the new measurement document type. Let me know if you see any issues!

stevevanhooser (approved):
Thanks, looks good! We can double check that our existing tutorials still work with these small changes to the docTables

stevevanhooser:
OK, so there is one conflict. It seems we both edited the openminds docTable method at some point. I'm not sure which is best. The pull request where I made the changes is here:

https://github.com/VH-Lab/NDI-matlab/pull/586

It seems that there is a condition in some datasets (can't remember which one) where an openminds document might have have as a property value an array of documents instead of a single entry. I am not zoomed in on this but I have a hunch that the one on main might handle all these cases.

Want to give that a try? (Am I making any sense? @jesshaley )

jesshaley:
In that case, I think we fixed the same bug. I’ll take a look at it in the
morning and check that main does the same thing and then let you know. If
so, we can merge with the main version. Thanks Steve!

jesshaley:
@stevevanhooser go ahead and merge with the main version!

PR #763, opened 2026-05-01
Add verbose logging to ndi.cloud.testLogin for debugging

stevevanhooser:
## Summary
Added comprehensive verbose logging to the `testLogin` function to help users debug authentication issues. When enabled, the function now prints step-by-step details about login attempts, token state, and any errors encountered.

## Key Changes
- Added `Verbose` parameter (default: false) to `testLogin` function arguments
- Implemented detailed logging throughout the login flow:
  - Initial state and configuration
  - Each of the three login attempts (probe, post-logout probe, post-UI-login probe)
  - Token state at key points (initial, after logout, after UI login, final)
- Enhanced `probe()` helper function with verbose logging:
  - API call details and results
  - Token verification steps
  - Error messages with identifiers
- Created new `reportTokenState()` helper function that displays:
  - Presence and length of `NDI_CLOUD_TOKEN` environment variable
  - Decoded JWT claims (email, sub, exp, iat)
  - Token expiration status and remaining validity
  - `NDI_CLOUD_ORGANIZATION_ID` environment variable state

## Implementation Details
- All verbose output is prefixed with `[testLogin]` for easy filtering
- Token state reporting safely handles missing tokens, decode errors, and expired tokens
- Token prefix (first 8 characters) is shown for identification without exposing full token
- Expiration times are displayed in local timezone for user convenience
- Error handling preserves original behavior while adding diagnostic information

https://claude.ai/code/session_01S6JAPmeYyWt6buMC3aUq6K


stevevanhooser:
@jesshaley Are you happy with this? I changed it so it has to have a good token to pass and it tries to obtain one if there isn't a good one. Previously if there was a token that didn't decode well it just passed it as not being able to check. But now it answers the question: "do we already have a good token?" and doesn't return true if we don't already have one. (It was asking previously, "if we have a token, is it still good?" and returned true if we didn't)

jesshaley (approved):
Works now!

Issue #805, opened 2026-05-31
ndi.cloud.sync.uploadNew discards upload return codes; sync index records documents whose binaries never uploaded


jesshaley:
## Summary

`ndi.cloud.sync.uploadNew` calls both `ndi.cloud.upload.uploadDocumentCollection` and `ndi.cloud.sync.internal.uploadFilesForDatasetDocuments`, then updates the sync index — but ignores the return values from both. When a binary upload silently skips or fails, the document is still recorded in the sync index as "uploaded", so:

- Sync reports success
- The cloud has the document but not the binary
- `ndi.cloud.api.files.listFiles` does not list the file UID
- A subsequent `Download → Export` call gets `"File not found"` from `ndi.cloud.api.files.getFileDetails` and the user sees a wall of 404s

Reproduces in nansen-pulakat against a cloud dataset where a sync round-trip completed without a binary upload running. See [Waltham-Data-Science/nansen-pulakat#43](https://github.com/Waltham-Data-Science/nansen-pulakat/issues/43) for the originating reproduction and diagnostic output.

## Smoking gun

`src/ndi/+ndi/+cloud/+sync/uploadNew.m` around lines 110-127:

```matlab
[b, uploadReport]=ndi.cloud.upload.uploadDocumentCollection(cloudDatasetId, documentsToUpload);
report.uploaded_documents_report = uploadReport;
if isfield(uploadReport, 'uploaded_document_ids')
report.uploaded_document_ids = string(uploadReport.uploaded_document_ids);
else
report.uploaded_document_ids = string(ndiIdsToUpload);
end

% Upload associated files
if syncOptions.Verbose
fprintf('Uploading associated data files...\n');
end
ndi.cloud.sync.internal.uploadFilesForDatasetDocuments( ...
cloudDatasetId, ...
ndiDataset, ...
documentsToUpload, ...
"Verbose", syncOptions.Verbose, ...
"FileUploadStrategy", syncOptions.FileUploadStrategy);
```
- `b` from `uploadDocumentCollection` is captured but never checked.
- `uploadFilesForDatasetDocuments` returns `[success, message]` per its own signature, but the call site discards both.
Then immediately after, `uploadNew` proceeds to step 5 (update the sync index) regardless of whether either upload actually succeeded.
## Concrete sequence that produces the bug
1. `uploadNew` is called by a downstream caller (pulakat's `Dataset → Methods → Sync` in this case).
[...]

PR #806, opened 2026-06-01
Fix sync index corruption when document or binary uploads fail

stevevanhooser:
## Summary
This PR fixes a critical issue where the sync index could be incorrectly updated to mark documents as "uploaded" even when their metadata or associated binary files failed to upload to the remote. This could leave the cloud in an inconsistent state with missing binaries, causing 404 errors in downstream operations with no way for automated retry to recover.

## Key Changes

- **uploadNew.m**: Added validation checks after document metadata and binary file uploads to prevent sync index updates if either upload fails. The function now returns early with an error message if uploads don't succeed, rather than proceeding to update the sync index.

- **filesNotYetUploaded.m**: Improved robustness when checking remote file upload status. If a remote file entry is missing the 'uploaded' field, the file is now conservatively re-queued for upload with a warning, rather than being assumed uploaded. This prevents silent data loss when API responses lack expected fields.

- **zipForUpload.m**: Added failure reporting for files that were skipped during upload because they were missing from disk. Previously these were silently ignored; now they cause the upload to be reported as failed, preventing documents from being marked as synced when their binaries weren't actually uploaded.

## Implementation Details

- Error handling is now consistent across metadata and binary uploads, with both preventing sync index updates on failure
- Conservative approach taken when upload status cannot be confirmed (missing fields or missing files)
- Verbose logging added to help diagnose upload failures
- Addresses issue #805 which tracked sync index corruption scenarios

https://claude.ai/code/session_01LTSt2kjigMbFf4pWmQaUVm


jesshaley (approved):
Reviewed and validated end-to-end on a real cloud dataset.

Test setup:
- Created a 100-byte test binary on disk
- Constructed a `generic_file` document referencing it via `session.newdocument(...)` + `doc.add_file('generic_file.ext', ...)`
- Ran `ndi.cloud.sync.uploadNew(dataset, 'Verbose', true)` on this branch

Result:
- `success=1` returned
- `Batch upload successful` in verbose output
- Sync index updated only after both uploads succeeded
- `uploaded_document_ids` contained the test doc's UID

Cloud-side confirmation via `getFileDetails(cloudDatasetId, fileUid)`:
ok=1
uid: '4126964cd4d5f4a6_40c958caa419e194'
uploaded: 1
size: 100
downloadUrl: <signed S3 url>

So `uploadNew` is now correctly propagating both upload return codes through to the sync index decision, and the batch upload pipeline through `zipForUpload` works end-to-end on the happy path. Fix is solid.
Unrelated finding while testing: `ndi.cloud.api.files.listFiles` errors with `Dot indexing is not supported for variables of type cell` at `ListFiles/execute:52` after a new file is uploaded. The response parser doesn't handle the heterogeneous shapes coming back from the API. Doesn't block this PR but worth its own ticket — I'll file separately if useful.

Issue #807, opened 2026-06-01
ndi.cloud.api.files.listFiles errors with "Dot indexing is not supported for variables of type cell" when the cloud response contains heterogeneous file entries


jesshaley:
## Summary

`ndi.cloud.api.implementation.files.ListFiles.execute` at line 51-52 assumes `dsetInfo.files` is a struct array, but `jsondecode` returns a cell array when the file entries have different field sets. This causes `dsetInfo.files(i)` to return a 1×1 cell whose `.uid` access fails with `Dot indexing is not supported for variables of type cell`.

In practice this is triggered any time a dataset's file list contains files whose attributes differ — for example after a freshly uploaded file lands alongside older entries that the API decorates differently. After the fix in #806 lets new uploads succeed, this latent listFiles bug surfaces on the next refresh.

## Reproduce

After running `ndi.cloud.sync.uploadNew` to upload a new file on top of an existing populated dataset:

```matlab
[ok, list] = ndi.cloud.api.files.listFiles(cloudDatasetId, "checkForUpdates", true);
```

Stack trace:

```
Dot indexing is not supported for variables of type cell.

Error in ndi.cloud.api.implementation.files.ListFiles/execute (line 52)
    if ~isKey(fileMap, file.uid)
                       ^^^^^^^^

Error in ndi.cloud.api.files.listFiles (line 56)
    [b, answer, apiResponse, apiURL] = api_call.execute();
```

## Root cause

`src/ndi/+ndi/+cloud/+api/+implementation/+files/ListFiles.m` lines 49-60:

```matlab
if ~isempty(dsetInfo) && isfield(dsetInfo, 'files') && ~isempty(dsetInfo.files)
    for i = 1:numel(dsetInfo.files)
        file = dsetInfo.files(i);
        if ~isKey(fileMap, file.uid)
            fileMap(file.uid) = struct(...
                'uid', file.uid, ...
                'uploaded', file.uploaded, ...
                'sourceDatasetId', file.sourceDatasetId, ...
                'size', file.size);
        end
    end
end
```

`jsondecode` returns `dsetInfo.files` as:
- a **struct array** when every JSON object in the array has the same field set
- a **cell array of structs** when JSON objects have different field sets

[...]

PR #808, opened 2026-06-01
Tolerate heterogeneous file entries in ListFiles.execute (#807)

jesshaley:
ndi.cloud.api.implementation.files.ListFiles.execute assumed dsetInfo.files is always a struct array and indexed it with (i). jsondecode actually returns a cell array when entries have different field sets, so (i) returns a 1x1 cell and the next line's file.uid errors with 'Dot indexing is not supported for variables of type cell'. The whole listFiles call dies.

PR #806 makes this reachable in practice: a fresh upload lands in the file list alongside older entries decorated slightly differently by the API, and the next listFiles call hits the cell-array branch.

Fix:
- Use {i} when dsetInfo.files is a cell, (i) when it is a struct array.
- Skip entries that aren't a struct or lack a uid so one bad entry can't take down the listing.
- Build the per-entry struct field-by-field with defaults so a missing optional field (uploaded, sourceDatasetId, size) is filled rather than crashing.

Verified original failure mode on R2026a after a fresh upload, and confirmed the new loop returns a complete listing without errors. See #807 for the reproduction trace."

### DIFF
--- a/src/ndi/+ndi/+cloud/+api/+implementation/+files/ListFiles.m
+++ b/src/ndi/+ndi/+cloud/+api/+implementation/+files/ListFiles.m
@@ -45,16 +45,38 @@ classdef ListFiles < ndi.cloud.api.call
                     return;
                 end
 
-                % Check if the 'files' field exists and is not empty
+                % Check if the 'files' field exists and is not empty.
+                % jsondecode returns dsetInfo.files as a struct array when
+                % every entry has the same field set, and as a cell array of
+                % structs when entries have heterogeneous field sets. The
+                % loop tolerates both shapes; missing optional fields on a
+                % single entry (uploaded / sourceDatasetId / size) are
+                % filled with defaults so one weirdly-shaped server response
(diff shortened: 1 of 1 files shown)


jesshaley:
Verified the fix against the failure mode that prompted the issue:
ok=1, count=13
Test file UID in list: 1
uid: '4126964cd4d5f4a6_40c958caa419e194'
uploaded: 1
sourceDatasetId: '69b1a69696e7e078981434b5'
size: 100

The pre-patch reproduction (a fresh upload on top of an existing
populated dataset) now lists all entries cleanly, including the
heterogeneously-shaped one that previously crashed the loop.

stevevanhooser:
Cool, will wait for CI to run, and will merge if it passes 
Thanks!

stevevanhooser (approved):

stevevanhooser:
@jesshaley OK for me to merge?

jesshaley:
@stevevanhooser , yes. please do. I'm finding additional issues, but will follow-up with those separately

Issue #151, opened 2020-06-26
hengen sorter 2 branch created


stevevanhooser:
Hi guys - 

Good news -- I merged the branches. Hengen sorter is now in the master.

The likely bad news is that I had to change a few things. 

First, ndi_experiment is gone and ndi_session is in its place. Second, ndi_thing is now ndi_element. Third, documents now have a "property_list_name" that indicates the name of the properties of that document. It was added for the validator, so the validator knows what field to look for to validate that document class. Each document class now only gets a single property list (although it can nest "dictionaries"/structs underneath this property list as much as it likes). I also changed the document class names so they are unique ("extraction_parameters"->"hengen_extraction_parameters" for example).

I made those changes in text to ndi_neuron_hengen.m and ndi_app_spikesorter_hengen.m and ndi_hengen_demo.m.

However, those changes haven't been tested. So I imagine that I probably broke some things with minor typos. The major logic should be there and working but maybe I broke something. 

Thanks!

stevevanhooser:
in progress

Issue #64, opened 2017-09-21
Terribly sorry accidentally pushed changes I didn't intend to


danielgmu:
So I tried to revert my commit but I'm not entirely sure everything is fine. Please let me know.

Issue #67, opened 2017-12-13
nsd_variable - add 'type' subfield?


stevevanhooser:
This would allow apps to find variables that they can work on by type rather than by name

PR #836, opened 2026-07-10
V_eta (Brainstorm J) migration: Path-S second pass

jesshaley:
## Summary

The NDI-matlab side of the V_eta / Brainstorm-J migration. Adds the **V_eta second pass** to `ndi.migrate.local` — the context-dependent fold that the single-document DID-matlab migrators (`+migrators_j`) cannot do because it needs the corpus-wide subject graph.

This is the counterpart to `DID-matlab` PR #151 (the `+migrators_j` migrators) and the `did-schema` V_eta tree / `V_eta_migration_plan.md` (PR #64).

## What it does

The pass-1 J migrators emit every anatomical site as a **located-by-default** `term_observation` (D3). This second pass promotes the ones that are **attributed** — the target of an intervention — to Path S:

- a **part-`subject`** (bare J identity),
- a **`term_assertion`** of its anatomical kind (the site ontology term), and
- a **`part_of` `directed_relation`** (part → animal),

and **retargets** every co-anchored manipulation on that animal onto the part. "Attributed" is decided conservatively from the graph: a site is promoted only when its source document also produced a manipulation (they share the migrated anchor id). A merely-located site (probe location, ontology label) is left as a value. Promotion is deduped per `(animal, site)` — the D3 mint allowlist realised as a graph predicate (discovery measured only 49 such loci, all Dab optogenetic, all distinct subjects).

## Changes

- **`+migrate/local.m`** — a V_eta branch in the second-pass step calls the new `resolvePathS`, which reads the migrated documents (`toStruct`), runs the promotion, and folds the minted bodies back through `v1_to_v2` (short-circuited as already-target) so they are padded/validated on the same footing as the deferral-resolution pass. Doc header + `Contents.m` updated.
- **`+migrate/+internal/pathSPromotion.m`** — the pure struct logic (find-or-create part-subjects, retarget, dedup), independent of any database or schema.
- **`tests/+ndi/+unittest/+migrate/TestPathSPromotion.m`** — unit tests for promotion, the merely-located skip, per-site dedup, and distinct-site separation (no database/schema needed).

## Notes

- The authoring environment has no MATLAB; validation runs via CI.
- The broader imageseries/dataseries/timeseries → data-type-leaf + `sampled_body` storage collapse is intentionally deferred to land with this NDI-side ingest work (per the `build_v_eta.py` header note). The `image_stack` fold itself (a body-backed `image_observation` + `sampled_body`) is in DID-matlab PR #151.

https://claude.ai/code/session_01BenWtpJyRu3QhEZqCErpwm

---
_Generated by [Claude Code](https://claude.ai/code/session_01BenWtpJyRu3QhEZqCErpwm)_


jesshaley:
## The 25 errors are ecosystem drift, not this commit — re-running won't clear them

Head `0aa80e41a` is **byte-identical** to the state that was fully green on 2026-08-25 (the three `test-eta-migrate*` push-runs all passed, and the Dab e2e was 7/0). It hasn't changed. So the `25 errors + 1 fail` that appear now cannot come from this commit's code — and `1 215 / 1 281` tests still pass, which rules out an infra/MatBox-install flake (that fails the whole suite, not a 25-test subset). Same code, green then and red now ⟹ the cause is **external drift** in what these tests validate against:

- **NDI `main` advanced** — this PR build merges against base `928b1cd5` (main gained ~11 templates + ~164 `.m` files, incl. the `spatialGeneExpression*` / `geneExpression` / `cellTypeLabels` / `fileReference` subsystem) since this branch was last green.
- **The DID-schema / DID-matlab V_eta siblings were restructured** by the `calculator-restructure` effort (DID-schema #67/#68/#69, DID-matlab #209) over 2026-09-21→22, changing the schema/migrator contracts the second-pass tests assert on.

**Re-running the job will not turn it green** — the failing subset is the V_eta second-pass tests, and they need the NDI second pass brought up to the restructured schema/migrator contracts and the advanced `main`. That's a code change requiring MATLAB (this authoring environment has none — validation only via CI) plus the restructure's design context, so it belongs with the coordinated restructure as it propagates to the NDI half (the DID-schema side is already green again on that effort's latest).

I can't re-run the check or make the MATLAB-side fix from here; flagging so the futile re-runs stop and the NDI update lands as part of the restructure. Keeping #836 watched until it's green.

---

Issue #862, opened 2026-07-27
Close PR #835 as obsolete — Java validator already removed on main (supersedes audit M3)


stevevanhooser:
## Summary

**PR #835 is obsolete and should be closed.** Its work — rebuilding the bundled Java validator JAR off the security-pinned `pom.xml` (the 2026-06 audit's "M3" gate) — has been made moot because the entire Java validator was subsequently **removed** from `main`. There is no longer a validator to build or harden.

## Background: what PR #835 set out to do

PR #835 ("Audit remediation (2026-06): rebuild the Java validator JAR off the security-pinned pom", by @audriB) was the last substantive gate of the 2026-06 NDI-matlab audit remediation. Its stated scope:

- Rebuild `src/ndi/java/ndi-validator-java/jar/ndi-validator-java.jar` off the M2 security-pinned `pom.xml`, so the shipped artifact matches the audited/pinned dependency set.
- Fix a pre-existing broken test surfaced by the rebuild (`EnumFormatValidatorTest`).

It was build/artifact + test only, with no runtime API change.

## Why it's now moot

Commit `4885b14` ("Remove unused Java schema validator (ndi.validate + ndi-validator-java)", 2026-06-15) removed the validator entirely, and its message explicitly notes it **supersedes the audit PR's M3**:

> ndi.validate (and the com.ndi/everit JAR it loads via checkJavaPath) is not on any runtime path: nothing in src/ or tests/ constructs ndi.validate, ndi.document.validate() is a stub that returns 1, and live document validation is handled by DID-matlab when documents are added to the database. The validator was legacy.
>
> This supersedes the audit PR's M3 (pom jitpack -> Maven Central) and makes the deferred 'rebuild the validator JAR' blocker and the GenBank-TSV/LFS item moot: there is no validator to build or harden.

That commit removed:

- `src/ndi/+ndi/validate.m`
- `src/ndi/java/ndi-validator-java/` (jar, `source-code/pom.xml`, docs, tutorials)
- `.github/workflows/build-validator-jar.yml`

## Confirmation on current `main`

The "unused" conclusion checks out on `main` today:

- `src/ndi/java/` now contains only an empty `.matlab2markdown-ignore` — the whole `ndi-validator-java/` tree is gone.
- `src/ndi/+ndi/validate.m` no longer exists. The remaining `validate` functions are unrelated (`ndi.cloud.sync.validate`, and the `ndi.document.validate()` stub that returns `1`).
[...]

PR #835, opened 2026-07-10
Audit remediation (2026-06): rebuild the Java validator JAR off the security-pinned pom

audriB:
Part of the 2026-06 cross-repo NDI ecosystem audit remediation (the NDI-matlab slice; M0–M6 landed previously). This is the remaining substantive gate.

## What
- Rebuilds the bundled Java validator **JAR** off the **security-pinned `pom.xml`** (the dependency pins from the M2 security pass), so the shipped artifact matches the audited/pinned build inputs.
- Fixes a **pre-existing broken test** surfaced by the rebuild.

## Why
The JAR in the tree predated the security pins, so the shipped validator didn't correspond to the audited dependency set. This makes the artifact reproducible from the pinned pom.

## Notes
- Build/artifact + test only — no runtime API change.
- Opened for your review; we are not self-merging.

PR #886, opened 2026-08-25
NDI-matlab — security/correctness proposals + symmetry counterparts (14 commits), JAR-free

audriB:
`audriB:catchup/ndi-matlab-2026-08` (`189bd7c29`) — **opened as a DRAFT against `main` only because we cannot create a base branch in your repo. First ask: please cut an integration branch (e.g. `integration/backend-hardening-2026-08`) from `main` and retarget this PR to it.** Opening against main is safe on the PR itself (fork PRs receive no secrets — the prod suite's jobs fail at authenticate instead of running; see "Why not main" below), but the merge should happen on the integration branch or after cherry-picking the CI-defusal commit to main. 14 commits (8 hardening + 4 symmetry counterparts + 2 schema fixes), **14 ahead / 0 behind** `origin/main` (`7509dc96c`).

## Standing framing — please read this first

**MATLAB is the source of truth. Every commit here is a conservative security/correctness proposal for your review, never architecture — and all of it is UNVERIFIED BY EXECUTION.** There is no MATLAB on the machine that produced this branch, so **zero tests were run**; every result below is a git-level fact (`ls-tree`, `merge-tree`, `rev-list`, `cat-file`) and is labelled as such.

## Why not `main` (P1)

`.github/workflows/test-cloud-prod.yml` on `origin/main` at `7509dc96c` still authenticates with `secrets.NDI_CLOUD_USERNAME` / `NDI_CLOUD_PASSWORD` against **production** and runs `testCloudApiProdPart(N, 3)` across a 3-way matrix, on `push` to `main` and on `pull_request` targeting `main`, with `cancel-in-progress: true`.

`pull_request.branches` filters on the **base** branch, so a PR targeting any branch other than `main` does not trigger it at all — targeting an integration branch is exactly sufficient. The sharper point: because we have `push=false` on this repo, our PR necessarily comes from a fork and GitHub does not pass repository secrets to fork `pull_request` runs, so a fork PR to main would burn 3 red jobs at `ndi.cloud.authenticate` rather than write to prod. **The real prod-write exposure is the `push` trigger on merge** — and with `cancel-in-progress: true`, a cancelled runner is SIGKILLed before MATLAB teardown, orphaning synthetic (possibly published) datasets in the public commons.

**This branch defuses that itself, and the fix can go to `main` first, safely.** `a79668910` (was `9c71fea30`) touches **only** `.github/workflows/test-cloud-prod.yml`, and both trigger blocks carry `paths-ignore` covering workflow files (`.github/workflows/**` on PR, `.github/**` on push). So a workflow-only PR carrying just that commit can target `main` directly: it will not trigger the prod suite on the PR, and will not trigger it on merge either. Cherry-pick it to `main` first and the trigger is defused for everything after. Zero commits have touched that file on main since our branch point, so it applies without conflict.

**One thing only you can check:** the commit's own comment notes that main's branch protection / required checks must stop requiring this workflow first. `GET /repos/VH-Lab/NDI-matlab/branches/main/protection` returns **404** for us — indistinguishable between "no protection configured" and "insufficient permission" (we are `admin=false`). Undetermined on our side.

## Symmetry counterparts (per your process guidance)

The last 6 commits add the MATLAB side of the paired symmetry batteries for the NDI-python catch-up PR (Waltham-Data-Science/NDI-python#61): `+makeArtifacts/+fun/{pathSafeName,whatVaries}.m`, `+readArtifacts/+fun/*`, `+readArtifacts/+session/objectTypeMarker.m`, the shared `+fun/cases.m` battery, and `FUN_CASES_SCHEMA.md` (the cross-language contract: canonical-string value grammar, join-by-name, allow-listed known divergences with an auditor). `+fun` is auto-discovered by `test-symmetry.yml` — no wiring needed. **Every one of these files is headed `/!\ AUTHORED WITHOUT A MATLAB RUNTIME` — your run is their first execution.** Two whatVaries cases are carried as predicted divergences (cell-valued constant parameters, all-NaN parameters — we believe `local_varyingFields`' `vlt.data.eqlen` errors on cells and reports all-NaN as varying; `runtests('ndi.unittest.fun.stimulus.whatVariesTest')` should settle it; if real, swapping `eqlen` for `isequaln` there matches `local_uniqueValues`).

## The 8 hardening commits (rebased onto `origin/main`)

| new | was | subject |
|---|---|---|
| `85ac14147` | `e56f5d746` | security(database): replace `eval` with validated `feval` in object reconstruction |
| `605416a26` | `1a8b41f77` | security(cloud): validate curl URL/path to block shell injection ← **conflict resolved here** |
| `ac2df649e` | `e63ef6a66` | security(cloud): sanitize server-supplied file uid to prevent path traversal |
| `493ca0d9f` | `a9059ae3d` | security(cloud): draw the AES-CBC IV from `SecureRandom` |
| `d06ef9cd5` | `73a81896b` | fix(time): repair dead-on-arrival syncgraph `removerule`/`removeepoch` |
| `29546720b` | `e4a203628` | fix(cloud): stop cloud sync reporting success when uploads failed |
| `a79668910` | `9c71fea30` | ci: stop prod-write cloud suite on push/PR; drop `cancel-in-progress` |
| `47d413f54` | `b4d67d71b` | ci: fail the no-cloud suite when zero tests discovered |

Each sink was re-checked against `main` as it stands **now**: `ndi_document2ndi_object.m:43` and `+epoch/+epochset/param.m:79` both still `eval`, and `validateClassName.m` is absent from main; `assertSafeCurlArgs.m` is absent from main and `GetFile.m:75` / `PutFiles.m:111` still interpolate server-supplied URLs into `system()`; `downloadDatasetFiles.m:67` still does `fullfile(targetFolder, file_uid)` with a server-supplied uid; `profile.m:418` still mints the AES-CBC IV from `int8(randi([-128,127],1,n))`.

## The JAR is gone

The branch's original root commit `1a5970250` re-added a **13,275,942-byte** validator JAR to a tree where you had already deleted the whole validator (`4885b1420`, 2026-06-15 — an *ancestor* of our branch point), and where it would have landed completely orphaned: no `+ndi/+validate`, no build workflow, no MATLAB caller. You closed **#835 as obsolete** on 2026-07-27 (tracked in #862). The rebase drops it mechanically.

Verified, not assumed: `git ls-tree -r --name-only HEAD | grep -i jar` → **exit 1, zero matches**, and the same grep against **each of the 8 intermediate commit trees** → zero matches, so the blob never exists at any point in the new history. `src/ndi/java/` in the final tree contains exactly one entry (`.matlab2markdown-ignore`), and no 13,275,942-byte blob appears anywhere. The diffstat arithmetic confirms a clean drop: 29 files +1236/−48 originally, `1a5970250` contributed 3 files / +360 → 26 files, +876/−48, binary line gone.

## Conflict resolution — 2 files, and a MATLAB parse is still owed

`merge-tree --write-tree origin/main catchup/…` → **exit 0, no conflict lines: CLEAN** (it was CONFLICT / 2 files before the rebase). 184 commits of upstream drift produced only 2 conflicting files: main touched 3 of the 29 files this branch touches, and `profile.m` moved three times and still auto-merges.

Both conflicts are in `executeWithCurl`, resolved keep-both: `assertSafeCurlArgs(...)` goes **above** main's `envPrefix = ndi.common.systemCurlEnvPrefix();`, and main's `LD_LIBRARY_PATH` comment block is retained verbatim. `PutFiles.m` critically keeps main's `[envPrefix sprintf([` opener so the pre-existing trailing `'"%s"'], this.filePath, this.preSignedURL)];` still balances. Zero conflict markers remain anywhere in `src/` or `tests/`.

**These two files have been verified only by (a) zero conflict markers tree-wide and (b) bracket/paren balance read by eye. A MATLAB parse of `GetFile.m` and `PutFiles.m` is the one thing still owed** — please run it before merging.

## SCIENTIFIC / needs your call: the `*_class` allow-list scan

`85ac14147` (eval→feval) means a real dataset carrying a `*_class` value **outside the `ndi.`/`did.` allow-list will now error on open instead of instantiating**. The commit itself asks for a read-only scan of `*_class` fields across existing datasets before this reaches main. That scan is yours, and it is a hard gate on the security half of this PR.

## Three NEW unvalidated curl sinks on `main` — not covered by this branch

Landed after our branch point, all reached through the new `ndi.common.systemCurlEnvPrefix` helper (which hardens `LD_LIBRARY_PATH`, not shell quoting):

- `src/ndi/+ndi/+database/+metadata_app/+fun/getRorId.m:4` — `baseUrl = "curl 'https://api.ror.org/organizations?query='"` then `sprintf('%s%s', baseUrl, name)`; the single quotes close **before** `name`, so the institution name is appended unquoted into the shell command.
- `src/ndi/+ndi/+database/+metadata_app/+fun/checkValidRORID.m:4` — `rorid` interpolated unquoted.
- `src/ndi/+ndi/+database/+metadata_ds_core/check_metadata_inputs.m:62` — `institutes{1}` interpolated unquoted.

These are metadata-editor fields rather than server-supplied, so lower severity than the cloud path, but they are a fresh regression of exactly the class this branch closes. Covering them needs `assertSafeCurlArgs` promoted out of `+cloud/+api/+implementation/+files` into a shared location — your call on placement, so it is not done here.

## V_eta interaction (#836)

**#836 keeps the eval sink and widens its input surface.** It replaces the inline field read with `obj_string = ndi.vintage.objectClass(ndi_document_obj, ndi_session_obj);` and leaves line 39 as `o = eval([obj_string '(ndi_session_obj, ndi_document_obj);']);`. Under V_eta the class name is resolved from a `software` entity behind a graph edge, so post-#836 the attacker-controllable path is *longer*, not shorter. The `validateClassName` + `feval` shape still applies verbatim — wrap the new call's return.

File overlap with #836 is 2 files (`ndi_document2ndi_object.m`, `+time/syncgraph.m`), both auto-merging, with disjoint syncgraph hunks (ours at L112/L560, V_eta's at L888). **Caveat: #836's head moved during our last session (`50bcd5731` → `1e3b89bfa`) and that overlap analysis was computed at the older head — re-run it before sequencing.**

## Not in this PR

- The 13 MB validator JAR (dropped; #835 closed as obsolete).
- The three new `main`-side curl sinks, and the promotion of `assertSafeCurlArgs` they require.
- Any V_eta / #836 reconciliation.
- Any architecture change — this branch is security/correctness only, per our standing rule.
- The `*_class` dataset scan and the `main` branch-protection question: both are yours.

---


audriB:
## First real MATLAB run of this branch — one red, ours, fixed

This branch had never been executed by MATLAB (authored without a runtime). `main`'s latest `Run NDI Test Suite No Cloud` is green, so the red was not pre-existing.

The good news first: of 873 tests in the no-cloud suite (867 passed / 1 failed / 5 incomplete), the blind-authored code produced **exactly one** failure. No parse errors anywhere in the resolved tree — the new `+fun` symmetry files, `cases.m`, and the GetFile/PutFiles conflict resolution all loaded and ran. The symmetry suite discovered 54 tests and ran 49 passed / 0 failed / 5 incomplete, with `makeArtifacts` a clean 16/16, so the `+fun` generators genuinely executed rather than being silently skipped.

**`ndi.unittest.time.SyncgraphMutationTest/testRemoveRuleDropsIndexedRule`** (fixed in `95bae27`):

```
verifyTrue failed.
--> The value must be logical. It is of type "double".
Actual Value: 1
```

Note the actual value: **1**. The rules *were* equal — the production `removerule` fix is correct and the test was wrong about types. `ndi.time.syncrule/eq` delegates to `vlt.data.eqlen`, which returns a double `0/1` rather than a logical, and `verifyTrue` rejects any non-logical actual value. Converted with `logical()` and gave the assertion a failure message; it still proves what it claimed (`removerule(1)` leaves `r2`, not `r1`).

I swept both the whole branch diff and the other new test files for the same two blind-authoring bug classes (non-logical fed to `verifyTrue`, and unqualified static calls on packaged classes) — this was the only remaining instance in this repo.

---

### One thing to know about the green `Test NDI Symmetry` badge

The no-cloud run reports **5 Incomplete** symmetry tests, all "Filtered by assumption":

```
ndi.symmetry.readArtifacts.fun.pathSafeName/testMatlabPythonSymmetry
ndi.symmetry.readArtifacts.fun.pathSafeName/testInputsAgree
ndi.symmetry.readArtifacts.fun.whatVaries/testMatlabPythonSymmetry
ndi.symmetry.readArtifacts.fun.whatVaries/testKnownDivergencesAreStillReal
ndi.symmetry.readArtifacts.time.timeConvert/testMatlabPythonSymmetry
```

The split is clean and worth stating precisely: for both `pathSafeName` and `whatVaries`, the `testMatlabArtifactsReproduce` leg **passed** — MATLAB really does re-read its own artifacts. Every skipped test is a *cross-language* one (`testMatlabPythonSymmetry`, `testInputsAgree`, `testKnownDivergencesAreStillReal`).

These do not fail the job — `Test NDI Symmetry` asserts only `all(~[results.Failed])` and allows Incomplete — so the badge is green while the cross-language leg has not run. Green here currently means "MATLAB can read its own output," not "MATLAB and Python agree." Notably, `testKnownDivergencesAreStillReal` is among the skipped: the test whose job is to catch a known divergence quietly disappearing is itself quietly absent.

This is the same issue DID-matlab's `d16b2ac` set out to fix, and the three repos in this catch-up program landed on three different conventions for it — DID fails on Incomplete (permanently red), NDR asserts `~Incomplete` but codes its python leg as an early `return` that scores as Passed (green), NDI allows Incomplete outright (green). Only DID is red, and its red is not less honest than the other two greens.

I have not changed any gate — picking one convention spans three repos and changes what green means, which is an owner call. Full comparison table is on DID-matlab#154.

PR #894, opened 2026-08-28
Supply the cloud file handler when adding documents, not just when opening them

stevevanhooser:
> **Update:** [VH-Lab/DID-matlab#156](https://github.com/VH-Lab/DID-matlab/pull/156) has merged, so the ordering blocker below is cleared. DID `main` now declares `customFileHandler` on `add_docs`.
>
> Note the current head is a `Update GitHub badges [skip ci]` commit from `github-actions[bot]`, so it had no CI runs of its own — the earlier failures were on `b11b170`. The three workflows have been re-triggered against this head via `workflow_dispatch`.

## Why

DID no longer downloads remote files itself. `did.implementations.sqlitedb.do_add_doc` previously called `ndi.cloud.api.files.getFile` directly, which made DID depend on NDI being on the path — a lower-level package reaching for a higher-level one. DID-matlab#156 removed that call in favour of the `customFileHandler` hook this class already uses at `open_doc`.

The consequence for NDI: a file location marked for ingestion that is not a local path now needs the handler passed at **add** time too, or it cannot be retrieved.

## What changed

- `do_add` passes `'customFileHandler', @download_file_from_cloud` to `add_docs`, mirroring what `do_openbinarydoc` already does at `open_doc`.
- `download_file_from_cloud` moves out of `do_openbinarydoc`, where it was a nested function and therefore unreachable from `do_add`, and becomes a file-local function. **Its behaviour is unchanged**: it resolves `ndic://<datasetId>/<fileUid>` by minting a fresh pre-signed URL through `getFileDetails` and downloading with `getFile`, and errors for any other location type.

Only `ndic://` is supported, as before — plain `http(s)` URLs are deliberately not handled by DID.

## Scope

Remote ingestion is expected to be rare: `ingest` defaults to 0 for `url` and `ndicloud` locations, so this path is reached only when a caller explicitly asks for a remote file to be ingested. Nothing else in NDI implements `do_add_doc` or subclasses `did.database`, so the added name-value pair affects nothing else.

## Testing

Not exercised locally — this session has no MATLAB. The change is a function move plus one added name-value argument, with no behavioural change to the handler itself. Coverage for the DID side of the contract is in DID-matlab#156 (`did.unittest.TestCustomFileHandler`, 7 tests); there is no NDI-side test for this path, which may be worth adding.


stevevanhooser:
## CI is red, and will stay red until [VH-Lab/DID-matlab#156](https://github.com/VH-Lab/DID-matlab/pull/156) merges

**Failing on `b11b170`:** `Run symmetry tests` and `Cloud CI 3`. (`Code Analyzer` passed; `Cloud CI 1`, `Cloud CI 2` and `Run tests (MATLAB latest)` were still running when this was written.)

The symmetry job reports 33 passed / 11 failed. Every failure is a `makeArtifacts` test that creates a session or dataset — `buildSession`, `buildDataset`, the four `blankSession*`, the three `ingestion*`, `rhdSeriesNavigator`, `timeConvert`. The `readArtifacts` counterparts are then skipped for missing artifacts. Tests that don't create anything (`preferences`, `profile`, `pathSafeName`, `whatVaries`, `downloadIngested`) all pass.

### Why

This is the cross-repo ordering dependency, confirmed rather than assumed:

1. `requirements.txt` pulls `https://github.com/VH-lab/DID-matlab` **unpinned**, so CI installs DID **main**.
2. DID main's `add_docs` declares only `OnDuplicate` and `Validate` in its `arguments` block, and an `arguments` block rejects unrecognized name-value pairs.
3. This PR makes `didsqlite.do_add` pass `'customFileHandler', @download_file_from_cloud` to `add_docs`.

So every document add fails, which is exactly the set of tests that break.

**Not pre-existing:** the symmetry workflow is green on NDI `main`, including run #702 on `5987a679` — the commit this branch is based on. This failure is this PR's.

### No fix to push here

The fix is DID-matlab#156, which adds `customFileHandler` to `add_docs`. It lives in the other repository, so there is nothing to port into this PR. Once #156 is on DID main, CI here should go green with no change to this branch.

I deliberately have **not** done either of the two things that would force green, because both would have to be reverted after #156 merges and both change shipped behavior for the sake of a CI artifact:

- pinning `requirements.txt` to a DID branch;
- guarding the `add_docs` call so it tolerates a DID that lacks the argument.

No re-run either — the failure is deterministic, not a flake.

### Merge order

**DID-matlab#156 first, then this PR.** I'll keep watching and will re-check once #156 lands.

---

stevewds (approved):

Issue #203, opened 2021-09-27
database gui: view of documents should show full human-readable json


stevevanhooser:
can use vlt.data.prettyjson to display the json contents

json_text = vlt.data.jsonencodenan(doc.document_properties);

json_text = vlt.data.prettyjson(json_text);

(can display that in the window)

PR #933, opened 2026-09-05
Lift the .gef/.h5ad ingestion path out of the GUI, and remember where a tile landed

stevevanhooser:
This branch carries **two independent pieces of work**. They do not depend on each other; they are together because they grew on the same branch. The larger half — the ingestion wrappers — was not described in the original version of this description, which covered only the tile memo. That is corrected here.

| | lines | what it is |
|---|---|---|
| **1. Ingestion wrappers** | ~890 | `.gef` and `.h5ad` → documents, callable without a GUI |
| **2. Tile path memo** | ~315 | `readViewport`/`exportRegion` stop re-resolving tiles they already found |

---

# 1. The `.gef` and `.h5ad` to documents path, lifted out of the GUI

## The problem

The only place that turned a Stereo-seq file into NDI documents was `ndi.gui.app.GeneIngest`. The sequence — read the file, make a geneList, make a pyramid, make the tiles, make the cells, make the labels — existed only as GUI callbacks. A script, a batch import, or the Python port had to either drive a GUI class or write the sequence again.

## The division it observes

**NDR reads the vendor's file and returns arrays; NDI turns arrays into documents.** Nothing in these functions parses HDF5, and nothing in NDR knows what a document is. `ndi.fun.doc.gene.fromGEF` and `fromCellBin` are the seam, and they exist because the seam was previously only crossed inside a GUI.

## `fromGEF`

`[pyrDoc, tileDocs, geneListDoc, info] = ndi.fun.doc.gene.fromGEF(session, gefPath, 'subjectID', id)` — the whole `.gef`-to-database path in one call, over `ndr.format.stereoseq.readGEF`.

**It reads the file once.** A real section is ~10⁸ records and minutes; the gene list and the pyramid both need those records, so they are read once rather than once per maker.

`chipSerial` and `basePixelSize` default from the file's own attributes. Note for anyone touching that conversion: `resolutionNm` is **nanometres** and `basePixelSize` is micrometres, so 500 nm → 0.5 — which is *also* `makePyramid`'s default, meaning a botched conversion looks correct on the common case.

`maxGenes` trims the read for a quick look. A trimmed pyramid is missing genes and nothing in the documents records that, so `info.notes` says so explicitly.

## `fromCellBin`

`[cellsDoc, labelDocs, info] = ndi.fun.doc.gene.fromCellBin(session, h5adPath, pyrDoc)` — the cells document positioned against the pyramid, plus one `cellTypeLabels` document per labeling asked for.

**It reads the file once, and that is the substance rather than a convenience.** `readCellBin` returns cells in file order and labels in file order, and the *only* thing tying a label array to a cell table is that they came from the same pass. Two passes that disagreed about row order — a column added, a filter applied, a different `obsColumns` request — would produce a label array that is still entirely valid and entirely wrong, on cells that are all real. A port that fetches labels in a separate call reintroduces exactly that hazard, and no test of either half would catch it.

**Labelings are asked for, never assumed.** A cellbin routinely carries a transferred atlas call and one or more unsupervised clusterings side by side; the file does not say which is which, and reading a cluster index as a cell type is a scientific error rather than a display bug. Pass nothing and you get cells with no labels. `info.labelColumns` lists what was available, with `readCellBin`'s name-based guess at each, so a caller can offer the choice rather than make it.

## `makeSourceFile`, and a dependency that was never populated

Both `spatialGeneExpressionTiles` and `spatialGeneExpressionCells` have declared a `source_file_id` dependency on `generic_file` since they were written. Nothing ever filled it. `makeSourceFile` creates that document and `fromGEF`/`fromCellBin` now set the dependency, so a pyramid can name the `.gef` it was built from.

**Describing is not attaching, and the default is to describe.** A SAW `.gef` is ~9.4 GB; ingesting a copy alongside the pyramid derived from it would roughly double the storage to hold bytes the pyramid already summarises. `attachFile` defaults false.

The checksum is the part worth having, and defaults on: a filename tells you what somebody called a file; a checksum tells you whether the file you have now is the file the pyramid was built from. Two SAW runs of the same chip produce files with the same name and different contents. Skipping it leaves the field **empty** rather than writing a placeholder that would later read as a real value.

## What changed in existing code

- `makePyramid.m`, `makeCells.m` — one optional `sourceFileID` argument each, defaulting to `''`. Behaviour is unchanged when unset.
- `GeneIngest.m` — now a thin adapter. `readNotes` became a forwarder to `ndi.fun.doc.gene.readNotes`; the cells step delegates to `fromCellBin`; the `labels` step is a **documented no-op**, because `fromCellBin` makes the label documents in the cells step from the same read. The step stays in the plan because the plan is what the confirmation screen describes, and a user should still see each labeling named before committing to it.

## Tests

`TestGeneFromFiles`, 19 cases, against the real NDR fixtures. They skip cleanly when NDR-matlab is absent — worth stating because that is how a green run could lie: CI reports **+27 tests with the skip count unchanged at 33**, so these genuinely ran rather than being filtered out.

---

# 2. Remember where a tile landed, in `readViewport` and `exportRegion`

## First, what is *not* wrong here

This started as a check for the eager-fetch bug found in NDI-python's `levelArrays`, which resolved every tile of every level while building the ladder — on a cloud session, downloading the whole pyramid before drawing a pixel.

**MATLAB does not have that bug.** `readViewport` and `exportRegion` already compute the intersecting tile range from the requested rectangle and open only those tiles, inside the loop. Neither is driven by a thread pool, so the SQLite-across-threads hazard that complicated the Python fix does not arise either. Nothing needed fixing on that front, and I would rather say so than manufacture a change.

## What does apply, and how much

Both readers open every tile they touch through `database_openbinarydoc`, on **every call**. `ndi.fun.doc.gene.tilePath` remembers the answer; a hit is one `isfile` check.

**Being precise about the size of this,** because the obvious justification is the wrong one. There is **no MATLAB viewer** for these pyramids — the only callers of `readViewport`/`exportRegion` are `readViewportBase` and the tests, and interactive viewing is napari's, through NDI-python. So this saves nothing for a viewer that pans, because there isn't one. (`ndi.gui.app.pyraview` is an electrophysiology signal viewer; the name is coincidental.)

What it does save is a **repeated programmatic read**: a sweep over overlapping rectangles, an export run for several gene subsets over one region, a montage. Each re-opens every tile it touches, and each open is a document read plus a location resolve. It is **not a download saved** — a cloud session's own cache already prevents a second retrieval. The win is modest and local.

The Python equivalent in [#203](https://github.com/Waltham-Data-Science/NDI-python/pull/203) is worth more, for a reason MATLAB does not share: napari re-renders on every gene toggle, and there the memo also keeps the fetch off a thread that may not touch the session.

## Why remembering is safe

The database names a cached file by its **immutable uid**, so the path for a given (document, filename) cannot change meaning: the file is either still there or gone, never different.

The key is document id **and** filename. Keying on the document alone would hand every tile of a level the first tile's bytes — a picture that renders perfectly and is wrong — so there is a test for exactly that.

## What the first CI run corrected

I had written that a vanished file "is fetched again", and a test asserting it. **That is only true of a cloud-backed session.** CI raised `DID:SQLITEDB:open` — *"The file tile.bin_0 in document ... cannot be accessed"* — and DID was right: on a directory-backed session the path **is** the stored file, not a cache copy, so a file that has gone is data that has gone and there is nothing to re-fetch from.

The help and the test now state the distinction and assert the invariant the two backings share: **whatever `tilePath` does with a path that has stopped resolving, it does not hand it back.** Trusting the memo would move the failure to whoever read the file, where a missing tile would look like a corrupt one.

Worth recording that this is the same shape as the bug the branch is about — behaviour identical on a directory-backed session, different on cloud, with only the cloud case costing anything. Reasoning about the local case and assuming the remote matched is what put the eager fetch in `levelArrays` to begin with.

## Tests

`TestTilePath`, 8 cases: the path is readable; the second call agrees with the first **and** with what the database returns directly; a vanished file is never handed back; the memo recovers once the file returns; two files of one document don't collide; `'clear'` empties it; missing arguments are named; and `readViewport` produces identical output through a warm memo and a cold one.

The memo is `persistent`, so it outlives a single test — every test clears it in setup and teardown, or they would read each other's entries.

---

# Status and what is not verified

CI is green: **12/12 checks**, 1,195 tests passing, both codecov gates passing with project coverage up 0.31%.

MATLAB is not runnable in the development environment, so **everything here executed for the first time in CI**. In particular:

- the wrappers have never been run against the real opossum `.gef`;
- there is no viewer check with a display;
- nothing here has been exercised against a **cloud-backed** session, which is the case the tile-memo half is reasoning about.

That last gap is the one worth naming loudly, because three separate mistakes on this branch came from assuming directory-backed behaviour matches cloud.

https://claude.ai/code/session_01YHxsZ56GZsZ3hJ98TJ81t9


stevevanhooser:
## `Cloud CI 1` is red on a test this PR cannot reach

```
ndi.unittest.cloud.FileSeriesRoundTripTest/testManifestSurvivesADownloadFromTheCloud
```

60 passed, 1 failed, 1 incomplete — **identical totals and the same single test** on both `a178e46` and `eae346a`. That is the one re-run, and it reproduced exactly, so this is not a flake being wished away.

### Why it is not this PR's

Checked rather than asserted:

- the commits on this branch touch **no cloud and no file_series file** — `git show --name-only` over them is four files: `fromGEF.m`, `GeneIngest.m`, and their two test classes;
- `FileSeriesRoundTripTest` was added by commit `023d6d6` ("Add demoNDISeries and the cloud round-trip acceptance test"), not here;
- that test file contains **zero** references to `gene`, `GeneIngest` or `fromGEF`.

It is a round-trip acceptance test against the live cloud, so it depends on service state this branch has no influence over. Worth noting alongside it that NDI-matlab `main` was independently red at `e200ab2` at the time of these runs, with the fix in flight as [#951](https://github.com/VH-Lab/NDI-matlab/pull/951); this branch's base predates that commit, so the two are not the same failure, but the cloud estate is clearly in motion this evening.

No fix for it exists that I could port into this PR, and inventing one here would mean widening a gene-ingest change into cloud file-series work. So this is a standing-down note rather than a patch: naming what is failing, and why it is not going to be fixed from this branch.

### What *was* mine, and is fixed

Separately, `Run tests (MATLAB latest)` failed on `a178e46` with **4 errors, 0 failures** — the four tests added for `stepArgs` and `gefProgress`. Both functions had landed in the `methods (Static, Access = private)` block, because the edit was anchored on `tick`, which is private; the tests could not reach them. `eae346a` moves both into the public `methods (Static)` block, where the rest of this class's pure, headlessly-testable layer already lives. Verified afterwards that all thirteen `GeneIngest` statics the test class calls now resolve to the public block.

The PR stays watched until that check confirms green.

---

stevevanhooser:
## Correction: a fix did exist, and merging `main` in resolved it

My [previous comment](https://github.com/VH-Lab/NDI-matlab/pull/933#issuecomment-5562176385) said of the `Cloud CI 1` failure:

> No fix for it exists that I could port into this PR

**That was wrong.** The fix was already on `main`. `Cloud CI 1` is green on `068a236`.

What I got right was that the failure was not caused by this diff. What I missed was checking whether the base had moved — and it had, by **34 commits**, almost all of them the file-series cloud work this very test exercises:

- `cd3c116` — reconstruct series `ingest_locations` on cloud download ([#958](https://github.com/VH-Lab/NDI-matlab/pull/958))
- `7473300` — upload file series members, not just the series manifest
- `262e505` — `extract_docs_files`: copy the series members, not just the manifest
- `9e59e3c` — batch presign in `download_file_from_cloud` ([#952](https://github.com/VH-Lab/NDI-matlab/pull/952))

and, decisively, `a878d30` — *"FileSeriesRoundTripTest: actually read the series back from the cloud"*. The test itself gained 225 lines on `main`. So this branch was running a **stale version of the test against stale download code**, which is exactly why it reproduced identically twice: it was deterministic, not flaky, and my re-run confirmed determinism rather than innocence.

The lesson for me is narrow and worth writing down: "this diff did not cause it" and "there is nothing this branch can do about it" are different claims. A failure in code the branch does not touch can still be a failure the branch's **stale base** is responsible for carrying. Checking `git rev-list --count <base>..origin/main` before standing down would have caught it immediately.

`068a236` is a clean merge of `main` (`5661afa`) — no conflicts, and the gene and `GeneIngest` files are byte-identical to `eae346a`.

### State on `068a236`

| Check | Result |
|---|---|
| Run tests (MATLAB latest) | ✅ 1,241 pass, 0 fail, 0 errors (1,274 total, 33 skipped) |
| **Cloud CI 1** | ✅ **now passing** |
| Run symmetry tests | ✅ |
| Code Analyzer | ✅ |
| codecov/patch, codecov/project | ✅ project 28.93%, **+0.33%** |
| Cloud CI 2, Cloud CI 3 | running |

Nothing outstanding on this PR that I am aware of.

---

stevewds (approved):

Issue #935, opened 2026-09-05
Cloud test suite: replace fatalAssert* with assert*/verify* so one failure doesn't abort the whole run


stevevanhooser:
## Summary

Most `ndi.unittest.cloud.*` test classes use `fatalAssertTrue` / `fatalAssertNotEmpty` / `fatalAssertNumElements` inside individual `Test` methods and inside `TestMethodSetup` / `TestClassSetup`. When any of these fires, MATLAB's `runtests` **aborts the whole run** — every subsequent test class in the suite is skipped, even ones that are completely independent.

Concrete example we hit today: an "Invalid or expired token" 403 from a stale cached token caused `testDocumentClassCounts/testResponseShape` to hit its `fatalAssertTrue`. The whole cloud test run aborted at that point:

```
Fatal assertion failed in ndi.unittest.cloud.testDocumentClassCounts/testResponseShape.
Test run aborted.
```

Everything after it (`testNdiQuery`, and any other tests that would have executed in that shard) never ran, so we had no signal on whether they'd have passed. And the failure that triggered the abort was itself transient — a rerun would have passed.

## Why `fatalAssert*` is (almost) never the right primitive in this suite

The `fatalAssert*` family is specifically for "the environment is so corrupted that continuing is unsafe or misleading" — e.g. the temp directory disappeared, the framework itself lost integrity. That's not what's happening in any of these calls; every one of them is checking a normal API result, a normal file state, a normal MATLAB return value.

For that shape of check, MATLAB's own guidance is:

- `verify*` — fail this method but keep the method running. Best for anything the test wants to keep asserting past, like multi-property comparisons of a returned struct.
- `assert*` — fail this method, stop the method, keep the class going. Best for "if this precondition didn't hold, the rest of the method is meaningless."
- `fatalAssert*` — abort the run. Reserved for "the framework is broken."

Almost every current use of `fatalAssert*` in these classes is really the "the precondition for the rest of this method failed" case, which is `assert*`, not `fatalAssert*`.

## Scope of the change

Files currently using `fatalAssert*` (from `grep -rn fatalAssert tests/+ndi/+unittest/+cloud/`):

- `DatasetDeleteAndUndeleteTest.m`
- `DatasetsTest.m`
- `DocumentsTest.m`
- `DownloadGenericFilesTest.m`
- `DuplicatesTest.m`
- `FilesDifficult.m`
- `FilesTest.m`
- `MatlabLicenseTest.m`
[...]

PR #937, opened 2026-09-05
compute: thread organizationId through startSession + tests (#936)

### LINKED ISSUE #936: ndi.cloud.api.compute.* should require an explicit organizationId; picking the first silently is wrong for multi-org users
## Summary

Every call under `ndi.cloud.api.compute.*` (and by extension, every `ndi.unittest.cloud.compute.*` test) fails with HTTP 400 for any user whose Cognito account is a member of more than one organization:

```
APIResponseBody: {
  "message": "Organization ID is required (user has multiple)"
}
```

The API is correct — it refuses to guess. The MATLAB client, however, silently picks the first org from the login response and never gives the caller a way to specify which org the compute session should run under. That works fine on the auto-testers configured in GitHub secrets (they belong to a single organization) so CI passes; it fails on any real user with multiple org memberships.

## Reproducer (from today, on prod)

Real error captured — abridged:

```
Running ndi.unittest.cloud.compute.ComputeTest
Warning: Login response contained 3 organizations; using the first
("649b1b1bea20f31db68d4f9f"). Selection among multiple organizations is not
yet supported.
> In ndi.cloud.authenticate>extractFirstOrganizationId (line 193)

Verification failed in ndi.unittest.cloud.compute.ComputeTest/testHelloWorldFlow.
  APICalledURL: "https://api.ndi-cloud.com/v1/compute/start"
  APIResponseBody:  { "message": "Organization ID is required (user has multiple)" }
```

Every test in the `compute` package fails the same way: `testHelloWorldFlow`, `testAbortZombieFlow`, `testZombieFlow`, and the follow-on errors (`sessionId` missing / `getSessionStatus` with empty id) all cascade from the same root cause. CI's canned account has one org and never sees this.

## Where the guess lives

`src/ndi/+ndi/+cloud/authenticate.m`, `extractFirstOrganizationId`:

```matlab
% Accept 1x1 struct, 1xN struct array, or cell array of structs.
...
if isstruct(orgs) && numel(orgs) >= 1 && isfield(orgs, 'id')
    organization_id = orgs(1).id;                  % <-- picks first, always
    nOrgs = numel(orgs);
...
% TODO: support selecting among multiple organizations explicitly.
if nOrgs > 1
    warning(...)                                    % noisy but harmless
end
```

The `organization_id` returned from here is what threads its way into every downstream API call — including compute — so any request whose backend requires an explicit org just inherits "the first one" without the caller ever knowing.

## What the compute commands look like today

`src/ndi/+ndi/+cloud/+api/+compute/`:

- `startSession(pipelineId, inputParameters)`
- `abortSession(sessionId)`
- `advanceSession(sessionId, ...)`
- `getSessionStatus(sessionId)`
- `listSessions()`
- `triggerStage(sessionId, ...)`

None of them accept an `organizationId` argument. `StartSession.execute` builds the request body from `pipelineId` + `inputParameters` and lets `authenticate` bind whatever org it decided. `listSessions` is similarly org-scoped by the token's implicit org.

## What we want

Make the org explicit on any compute command whose backend behavior depends on it — at minimum `startSession` and `listSessions`, and worth reviewing for the others while we're in there. Concretely:

1. **`startSession` gains a required `organizationId` argument.** Not optional-with-default — required. The compute session belongs to an org, that org bills the run, and asking "which org?" is not something the client should be guessing. Signature:

    ```matlab
    function [b, answer, apiResponse, apiURL] = startSession(organizationId, pipelineId, inputParameters)
    ```

    The API request body then includes it explicitly, so the backend never has to disambiguate.

2. **`listSessions` gains a required `organizationId` argument.** Sessions are org-scoped; listing across orgs isn't a thing the API supports today and shouldn't be silently emulated by the client.

3. **`authenticate` stops guessing.** Downgrade the "picks the first, warns" behaviour to a hard error (`NDI:Cloud:AmbiguousOrganization`) when the user has multiple orgs, with a message that names the orgs available and points at the new required argument on compute commands. Callers whose flow doesn't need org disambiguation (single-org users, non-compute endpoints where the API doesn't require it) stay working.

    Or, if that's too aggressive to land in one PR: keep the warning but stop propagating the guessed id anywhere downstream — pass it explicitly at each call site that actually needs it, and let the sites that don't need it not receive it.

4. **`ndi.unittest.cloud.compute.*` tests are updated** to pass `organizationId` explicitly, using a new env var (`NDI_CLOUD_TEST_ORGANIZATION_ID`) with a `assumeNotEmpty` at setup. Auto-testers configured in GitHub secrets get the single org they belong to; developers with multiple orgs export the one they want to use. CI keeps working; local runs stop erroring.

## Why it matters

Any NDI user with membership in more than one organization — a lab manager who belongs to multiple sites, a collaborator with a personal org plus a shared one, a contractor with client orgs — cannot run the compute pipeline from MATLAB without editing the code. The error message they see is server-side ("Organization ID is required") and doesn't point at the fix. The current warning in `authenticate` is not surfaced by tests running under `runtests`, so a user can quite reasonably wonder if the API is broken.

Making the argument explicit and required is what the API already expects; the client should match.

## Suggested acceptance criteria

- `ndi.cloud.api.compute.startSession` requires `organizationId`; calling it without one is a compile-time error at the `arguments` block.
- `ndi.cloud.api.compute.listSessions` requires `organizationId`.
- `authenticate` no longer silently picks an org; either errors on ambiguity or stops feeding a guessed value into downstream calls.
- `ndi.unittest.cloud.compute.*` tests read `NDI_CLOUD_TEST_ORGANIZATION_ID` (or equivalent), `assumeNotEmpty` in setup, and pass it through explicitly.
- Nightly CI on the current secret account keeps passing (its account has one org).
- A user with multiple orgs, with `NDI_CLOUD_TEST_ORGANIZATION_ID` set to any of their org ids, can run the compute tests locally and see them pass.

Filed from investigation on VH-Lab/NDI-matlab#934 while working through cloud test failures.

stevevanhooser:
## Summary

Every `ndi.unittest.cloud.compute.*` test fails today for any user who belongs to more than one organization:

```
APIResponseBody: {
  "message": "Organization ID is required (user has multiple)"
}
```

The API's compute controller requires `organizationId` in the POST body of `/compute/start` whenever the caller has more than one org; if not supplied it 400s. The MATLAB client was letting `ndi.cloud.authenticate` silently pick the first org from the login response and never plumbed it into the compute calls, so the API never received it. Auto-tester accounts in GitHub secrets belong to one org and never saw the failure; real developer accounts with multiple orgs hit it every time.

Closes VH-Lab/NDI-matlab#936.

## Scope

Narrowed from the issue: only `startSession` needs the change. `listSessions` on the API side (`api/src/controllers/compute.controller.ts` line ~363) uses `computeRepository.listByUser(userId, dateFilter)` — filtered by user id, not org — and already works for both single- and multi-org accounts.

Every other command (`abortSession`, `advanceSession`, `getSessionStatus`, `triggerStage`) is keyed by `sessionId`, which the API already scopes to the caller's user; they don't need `organizationId` either.

## Changes

**Client:**

- `ndi.cloud.api.compute.startSession(pipelineId, organizationId, inputParameters)` — gains a required positional `organizationId` (2nd arg; `inputParameters` stays 3rd and still optional). Callers who pass empty (`""`) get the pre-existing "let the backend pick" behaviour, so single-org accounts keep working; multi-org accounts must now pass their id explicitly.
- Implementation writes `organizationId` to the request body only when non-empty.
- `ndi.cloud.helloMatlab` — new Name-Value `OrganizationId` option, default `""`. Passed through to `startSession`.

**Tests:**

- `ComputeTest`, `ZombieTest`, `HelloMatlabTest` — read `NDI_CLOUD_TEST_ORGANIZATION_ID` from the environment at `TestClassSetup` and thread it through every `startSession` / `helloMatlab` call.
- Single-org accounts (CI's current secret) can leave the env var unset; multi-org accounts set it once and every compute test in the class picks it up.
- Downgraded the pre-existing `fatalAssertNotEmpty` on credentials to `assumeNotEmpty` in these classes, matching the guidance in VH-Lab/NDI-matlab#935 (one failure should not abort the whole run).

**Workflows:**

- `.github/workflows/test-cloud-*.yml` — plumb the new `NDI_CLOUD_TEST_ORGANIZATION_ID` secret through. An unset secret arrives as an empty string, so this is a no-op for the current single-org CI account and future-proofs the setup if the account is ever replaced with a multi-org one.

## Test plan

- [ ] Local run against a multi-org account with `NDI_CLOUD_TEST_ORGANIZATION_ID=<some org>` set — `runtests('ndi.unittest.cloud.compute')` passes end-to-end (or in `testHelloWorldFlow`'s case, at least reaches the actual `startSession` response instead of 400).
- [ ] Nightly CI on the current secret single-org account keeps passing (`NDI_CLOUD_TEST_ORGANIZATION_ID` unset → empty → backend auto-picks).
- [ ] `ndi.cloud.helloMatlab('OrganizationId', someOrg)` accepts and forwards the value; no positional/name-value collision.

https://claude.ai/code/session_014pMyfBKMMch9wVNMEcYfug

---
_Generated by [Claude Code](https://claude.ai/code/session_014pMyfBKMMch9wVNMEcYfug)_


stevevanhooser:
Both Codecov checks (`codecov/project` −0.01%, `codecov/patch` 0.00% of diff hit) are red on the current head, and neither is this PR's to fix:

- The diff only touches cloud-API wrappers, `helloMatlab`, and the `+cloud/+compute/*` integration tests. Those paths are only exercised by `test-cloud-prod.yml` / `test-cloud-api.yml`, which don't upload to Codecov.
- The `-0.01%` project drop is the arithmetic effect of the added-but-unmeasured lines, not a real coverage regression.

The actual functional CI (Cloud CI 1/2/3 on `test-cloud-prod.yml`, PR-run #510) reported 1059 passed / 0 failed vs. base. Ready for review.

---

stevewds (approved):

Issue #178, opened 2021-05-27
Discussion of Tutorial 2.3


stevevanhooser:
Enter your comments, questions, bugs, or discussion items for Tutorial 2.3 here.

KellyZhangL:
Codeblock 2.3.2.1 `p = S.getprobes('type','n-trode');` returns a _0x0 cell_
This later prompts the below error when I run `se.extract(p{1},[],my_extraction_name{1},redo);` 

> Index exceeds the number of array elements (0).

KellyZhangL:
I believe the above error was caused by running `S.database_clear('yes');` to see if I could resolve a "File not found or you do not have permission to access this" error when running `se.extract(p{1},[],my_extraction_name{1},redo);`

I retyped code from Tutorials 2.1 and 2.2 to see if I could add something that would help with p = 0 probes (p resulted in 2 probes this time, both I believe are from the 2.1 Tutorial). I am getting the below error again when I run `se.extract(p{1},[],my_extraction_name{1},redo);`

> Warning: File not found or permission denied

stevevanhooser:
Hi Kelly -

If you run `S.database_clear('yes');` then you'll need to start back at Code block 2.3.2.1

Also, instead of opening the session with `ndi.session.dir`, you'd need to re-add the daq systems. I changed the tutorial so that code block 2.3.2.1 now uses `ndi.setup.vhlab` instead of `ndi.session.dir`, which should re-build the basic daq systems.

So, please try from Code block 2.3.2.1 and if you're still getting errors we will get to the bottom of them!

Thanks
Steve

KellyZhangL:
I tried what you suggested but I'm still getting the same "Warning: File not found or permission denied" error after `se.extract(p{1},[],my_extraction_name{1},redo);`.

stevevanhooser:
Hmm. OK, seems like this is what we should troubleshoot. Does it give you an error or just this warning? Can you paste in the complete warning/error stack? (usually it will tell you which function produced the warning).

Tomorrow we can zoom about it if we can't solve it here...

KellyZhangL:
Warning: File not found or permission denied 
> In release_lock_file (line 48)
In vlt.file/dumbjsondb/closebinaryfile (line 346)
In ndi.database.implementations.database/matlabdumbjsondb/do_closebinarydoc (line 109)
In ndi/database/closebinarydoc (line 124)
In ndi/session/database_closebinarydoc (line 361)
[...]

Issue #177, opened 2021-05-27
Discussion of Tutorial 2.2


stevevanhooser:
Enter your comments, questions, bugs, or discussion items for Tutorial 2.2 here.

PR #939, opened 2026-09-06
Signed URL set client in ndi.cloud.api.documents

stevevanhooser:
Wraps the `signed-url-set` endpoints in `ndi.cloud.api.documents`. This is the download side of the file-series work — the piece that lets a viewer resolve a uid to a URL without a round trip per tile.

Independent of the DID file-series PRs: NDI pulls DID from `main` (per `requirements.txt`), so nothing here calls the new DID API. It can merge on its own.

## What it wraps

Both server forms:

| function | endpoint |
|---|---|
| `getSignedURLSet` | `GET .../signed-url-set` — one page |
| `getSignedURLSetAll` | the same, following `nextCursor` to the end and merging |
| `createSignedURLSetJob` | `POST .../signed-url-set-jobs` — 202 with a `jobId` |
| `getSignedURLSetJobStatus` | `GET /signed-url-set-jobs/{jobId}` |
| `waitForSignedURLSetJob` | polls the above with exponential backoff |
| `getSignedURLSetResult` | downloads and parses the gzipped result blob |

Each follows the existing wrapper/implementation-class split, returns the usual `[b, answer, apiResponse, apiURL]`, and `waitForSignedURLSetJob` mirrors `ndi.cloud.api.implementation.files.WaitForBulkUpload` directly.

Every entry point takes an optional `fileSeries` name, so a document holding a dual image pyramid can ask for the one level being viewed rather than all of it.

## The uid recovery

This is the part worth reading, and the reason these functions return a `containers.Map` rather than the raw decoded struct that every other call in this package returns.

A `did.ido` uid is `NUM2HEX(<serial date>) '_' NUM2HEX(<random>)` — 16 hex digits, an underscore, 16 more. A MATLAB struct field name cannot begin with a digit, so `jsondecode` silently renames those fields, prefixing an `x`. Reading uids back off the decoded struct would hand callers uids that do not exist. In a sample of 2000 realistic uids, **1274 began with a digit**.

Field *order*, however, survives decoding. So `signedURLFileMap` takes the values from the decoded struct — letting `jsondecode` handle string escaping inside the URLs — and the keys from the raw payload, which still has the uids verbatim, and zips them. If the two counts disagree it errors rather than pairing them anyway: a wrong uid downloads the wrong file, which is worse than a failed call.

The raw payload is available because these requests are sent with `HTTPOptions('SavePayload', true)`.

## The payload scan

`signedURLFileMap_keys` is vectorized rather than a character loop. A pyramid level's signed URL set is several MB of JSON, and an interpreted per-character walk over that is seconds of wall time per page.

It is a real JSON scan and not a regular expression — backslash escapes and string state are tracked — so none of these can break it:

- a `{` inside a signed URL value
- an escaped `\"` inside a value
- the literal text `"files"` appearing as a value
- a nested `files` object under some other key
- the `files` object appearing after other keys

I verified the algorithm by porting it to numpy and running it against those cases plus a 2000-entry payload before writing the MATLAB, since there is no MATLAB in this environment. All 14 cases and the large payload agreed with the expected key list; the MATLAB tests below mirror them.

One deliberate change during that pass: an earlier draft used `discretize` with `±Inf` edges, which I could not verify here. It is now a `cummin` lookup, which is plainly correct and needs no such assumption.

## Gzip

`getSignedURLSetResult` downloads through `ndi.cloud.api.files.getFile` rather than `websave`. That path defaults to curl, which requests identity encoding so the gateway does not hand back a body `websave` would auto-inflate and corrupt — exactly the hazard that made `getFile` prefer curl in the first place, and this blob is gzipped. It then decides whether to decompress from the gzip magic number rather than assuming either way.

## Tests

`tests/+ndi/+unittest/SignedURLSetTest.m`, 22 tests, all offline. They live outside `tests/+ndi/+unittest/+cloud` deliberately: `testToolboxNoCloud()` excludes that namespace by prefix, so tests placed there would not run in the no-cloud CI job. Covered: endpoint construction, the missing-path-parameter error, query-string assembly and cursor encoding, and the uid recovery.

One test asserts that `jsondecode` really does rename a digit-leading field, so the suite still means something if that behaviour ever changes rather than quietly testing nothing.

## Server-side prerequisite

The endpoint builds its uid set from the files a document references, which today means `file_info` entries. **File series members live in the manifest blob, not in `file_info`**, so `ndi-cloud-node` has to learn about `file_series` before these calls return series members, and before the `fileSeries` filter does anything.

This corrects a conclusion from the earlier scoping: "ndi-cloud-node needs no change" was about *upload*, where the bulk zip path is uid-keyed and genuinely needs nothing. It does not carry to the download side, which has to enumerate a document's uids. Everything in this PR is additive and inert until the server knows, so it does not need to wait.

Co-Authored-By: Claude Opus 5 &lt;noreply@anthropic.com&gt;

https://claude.ai/code/session_0186NbFn61EbbPTYZgVhdAoE


stevevanhooser:
## Correction, and the design call it opens

The commit message on 023d6d6 says `ndi.document`'s duplication of DID's file API is "forced rather than incidental." **That is overstated.** Correcting it here because it changes what should happen next.

What is true: `did.document`'s constructor calls `did.document.readblankdefinition` with the class name written out, and MATLAB does not dispatch an explicitly qualified static call virtually. NDI's resolver differs — it registers `$NDISCHEMAPATH` through `ndi.common.PathConstants.updateDIDConstants` when DID has not seen it, and recurses into its own resolver for superclasses.

What I missed: `did.document` also accepts a **pre-resolved properties struct**. So a subclass could resolve with NDI's resolver and hand the result up:

```matlab
function obj = document(document_type, options)
    props = ndi.document.readblankdefinition(document_type);   % NDI's resolver
    obj@did.document(props);                                   % struct path, skips DID's
    ...
end
```

That sidesteps the static-call problem entirely. So the duplication is friction, not a blocker — most likely historical.

I also checked the handle/value question, since that *would* be a hard blocker: MATLAB forbids a handle class subclassing a value class or vice versa. Both `did.document` and `ndi.document` are value classes (`classdef document`, neither `< handle`), so that constraint does not apply either.

### The call

File series therefore needs one of two things in `ndi.document`, and it is worth deciding before writing either:

1. **Mirror the series methods** — `seriesNames`, `isFileSeries`, `seriesCount`, `seriesSourceRoot`, `seriesIngestLocations`, `addFileSeries`, `removeFileSeries`, the `NAME_#` → `file_series` rule in `is_in_file_list`, the `add_file` refusal, and the declaration-level exclusivity validation. That is roughly 460 lines duplicated from VH-Lab/DID-matlab#178, plus its tests, plus both copies to keep in step afterwards.

2. **Make `ndi.document` subclass `did.document`** via the struct-constructor route above. Larger and riskier now, but it retires the whole class of drift rather than adding to it — and `current_file_list` not expanding a series is exactly the kind of rule that would otherwise be fixed in one copy and not the other.

I have not done either. The acceptance test in 023d6d6 asserts on the behaviour, not the mechanism, so it works unchanged under either.

Worth noting what is *not* blocked: the manifest reader/writer is already in DID `main` (#177), so `ndi.document.addFileSeries` — mirrored or inherited — does not have to wait for #178 to merge.

---

stevewds (approved):

PR #944, opened 2026-09-06
Fix #945: stop destroying files.series_info after download, and test the round trip that found it

### LINKED ISSUE #945: updateFileInfoForLocalFiles destroys files.series_info: reset_file_info clears it and only file_info is repopulated
A document with a file series comes back from the cloud still **declaring** the series but reporting **zero members**. The series name survives; `count` and `n_present` come back as `0`.

Found by `ndi.unittest.cloud.FileSeriesRoundTripTest/testManifestSurvivesADownloadFromTheCloud` on VH-Lab/NDI-matlab#944, which is the first test that reads a series back out of the cloud rather than out of the local dataset it was built in.

## What the run shows

Upload of a 4-member series succeeds, the download succeeds, the document is found — and then:

```
Verification failed ... downloaded series declares a different member count
    Actual Value:   0
    Expected Value: 4
Verification failed ... downloaded series lost members from its manifest slots
    Actual Value:   0
    Expected Value: 4
```

`isFileSeries('chunkdata.bin')` **passes** on the downloaded document, so `series_info` survives with its name intact. `seriesCount` reads `series_info(index).count` / `.n_present` directly, and it returned a literal `0` rather than erroring or returning empty — so the field is present and its value is zero, not missing.

## Why this is a defect and not the expected stored form

Stripping on the way to storage is supposed to remove member paths **only**. `did.document.stripSeriesIngestLocations` empties `ingest_locations` and nothing else, and DID's own test states the contract:

```matlab
testCase.verifyEmpty(stored.seriesIngestLocations('chunkdata.bin'), ...
    'a stored document no longer says where its members came from');
testCase.verifyEqual(stored.seriesCount('chunkdata.bin'), 1, ...
    'but it still knows the series and its size');
```
— `did.unittest.TestDocumentFileSeries/testAccessorIsEmptyOnAStoredDocument`

And `seriesCount`'s docstring explains why the number is kept on the document at all:

> Without it, answering "how many members?" would mean either paging a cloud listFiles or fetching the manifest — which for a large series is most of a megabyte, possibly over the network. A count is a few bytes and is wanted constantly: sizing an upload, reporting progress, deciding whether a dataset is complete.

That is exactly the guarantee this breaks. On a downloaded dataset the cheap answer is now wrong, and it is wrong **silently** — `isFileSeries` still says yes, so a caller sizing an upload or reporting progress gets `0` for a 28,000-member series rather than an error.

## Where it is not

Two paths preserve the counts, which narrows it:

- **strip + rebuild**, per the DID test quoted above
- **local database round trip** — `testDocumentReportsItsSeries` does a `database_search` against the local dataset and gets 4, in the same run, minutes earlier

So the loss is somewhere in upload → server → download. `ndi.cloud.upload.uploadDocumentCollection` ships the whole property struct (`jsonencodenan(docProperties)`), and `ndi.cloud.download.downloadDocumentCollection` is generic (`rehydrateJSONNanNull` → `jsondecode` → `structsToNdiDocuments`) with nothing series-aware in it. Neither has any `series_info` handling at all. That leaves, roughly in order of suspicion:

1. the server persisting documents against a schema that does not know `files.series_info` and defaulting its numeric fields
2. `jsonencodenan` / `rehydrateJSONNanNull` round-tripping the nested `series_info` struct lossily
3. `jsondecode` reshaping `series_info` when `ingest_locations` is an empty struct array

I could not distinguish these without inspecting the uploaded payload and the stored server record, which needs someone who can see both.

## Reproducing

`ndi.unittest.cloud.FileSeriesRoundTripTest/testManifestSurvivesADownloadFromTheCloud` on the branch for #944. Observed against **prod**. Whether the newer dev API behaves the same is not yet known — that is worth checking as part of triage, since it would say whether this is server-side and already addressed there.

## Note on the manifest itself

The manifest *file* appears to survive — the failures are on the document's own count fields, not on reading the manifest back. The later assertions in that test (manifest `count`, and uids matching slot for slot) did not report failures, so the bytes made the journey. It is the cheap metadata beside them that did not.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

https://claude.ai/code/session_0186NbFn61EbbPTYZgVhdAoE

---
_Generated by [Claude Code](https://claude.ai/code/session_0186NbFn61EbbPTYZgVhdAoE)_

stevevanhooser:
This started as a test-only PR. The test found a real bug on its first run, so it now carries the fix too.

## The bug (#945)

`ndi.cloud.sync.internal.updateFileInfoForLocalFiles` runs on every downloaded document when `SyncFiles` is true, to repoint file locations at the freshly downloaded copies:

```matlab
originalFileInfo = document.document_properties.files.file_info;
document = document.reset_file_info();          % clears file_info AND series_info
for i = 1:numel(originalFileInfo)
    ... document = document.add_file(filename, file_location);   % restores file_info ONLY
end
```

`did.document/reset_file_info` clears `files.series_info` along with `files.file_info` — its own comment says *"A series' per-instance record is reset with the rest"*. The loop repopulates `file_info` only. This function predates file series.

So a downloaded document reported **zero members for a populated series**, and did it silently: `isFileSeries` reads `files.file_series`, the class declaration from the schema, which the reset deliberately leaves alone. A caller sizing an upload or reporting progress on a downloaded 28,000-member series was told "yes, this is a series" and "it has 0 members", with no error anywhere.

Only the `SyncFiles = true` branch is affected (`downloadNdiDocuments` line 125). `updateFileInfoForRemoteFiles` (line 131) mutates `file_info` in place and never resets, so it was never affected.

## The fix

Capture `series_info` before the reset, restore it after the loop with `setproperties`, matching how the remote-side twin writes `files.file_info` back. Carried verbatim: the record holds `name`, `count`, `n_present` and `source_root`; its `ingest_locations` were already emptied on the way into storage, and nothing on this path re-ingests members, so there is nothing to recompute.

## How it was found, and a diagnosis I got wrong

`testManifestSurvivesADownloadFromTheCloud` is new here. The three tests that already existed in `FileSeriesRoundTripTest` upload the dataset and then assert against `testCase.LocalDataset`, which is never re-downloaded — so they established that the cloud *accepts* a document carrying a series manifest, and nothing more. A server that dropped, renamed or altered the manifest would have passed all three. This one downloads the dataset into a fresh folder and reads the series back out.

It failed immediately, and **I diagnosed it as a server bug and was wrong**. The full retraction is on #945. The short version: the downloaded `file_info` I cited as having "survived intact" had `"location": "/tmp/download/files/…"` — a local download path. It had not survived; it had been rebuilt by the very loop that was clearing `series_info`. I built an argument on the contrast between a rewritten field and a cleared one without noticing either was which. Credit to the ndi-cloud-node reviewer who said the server could not be doing this.

## What is here

**Fix**
- `updateFileInfoForLocalFiles` preserves `series_info` across the reset

**Tests**
- `TestUpdateFileInfoPreservesSeries` (4) — calls the function directly with a local directory standing in for the download folder. No credentials, no network. The regression test for this fix, and it also pins *why* the loss is silent, and pins the repointing behaviour I misread.
- `TestDocumentSeriesJsonRoundTrip` (5) — the no-network JSON encode/decode round trip, including the batch cell-array encoding `zip_documents_for_upload` actually performs. Written to split client-from-server; it exonerated the client, which in hindsight should have pointed me back at post-download mutation sooner.
- `testManifestSurvivesADownloadFromTheCloud` — the cloud round trip that found the bug. Now passes.
- `FileSeriesRoundTripTest` now carries the narrative + `ndi.unittest.cloud.APIMessage` pattern used by every other suite in `tests/+ndi/+unittest/+cloud`. It was the outlier. The failure that produced the final diagnosis was legible only because of this — it prints whether `files.series_info` is present and dumps the whole `files` struct.

## Verification

All 12 checks green at `045c0ee`:

| check | result |
|---|---|
| Run tests (No Cloud) | **1,210 tests / 192 suites / 1,177 passed / 33 skipped / 0 failed** |
| Run symmetry tests | pass |
| Cloud CI 1, 2, 3 | pass — Cloud CI 1 is the one that was red |
| Code Analyzer | pass |
| codecov/patch, codecov/project | pass |

Codecov: Lines +5, **Hits +14, Misses −9**. The fix is five lines and turned nine previously-unexercised lines into covered ones — `updateFileInfoForLocalFiles` had no direct test before this, which is a fair part of how it survived.

Also confirmed by @stevevanhooser locally against the **dev** API, where the bug reproduced identically before the fix.

## Still skipped, correctly

`testMembersSurviveTheRoundTrip` remains Incomplete. Its gate is local — `database_existbinarydoc(doc.id(), 'chunkdata.bin_1')` — and members are never ingested: `addFileSeries` records where they are, nothing copies them into `FileDir/<uid>`, and nothing resolves `chunkdata.bin_1` to a uid on read. That is VH-Lab/DID-matlab#173, now being worked separately. When it lands, this test flips to passing with no edit.

## Related

- #945 — the bug, retitled and retracted from its server-side framing
- #946 — the same `reset_file_info` pattern in `ndi.database.fun.extract_docs_files`, which returns documents that `ndi.dataset` stores. Separate PR.
- VH-Lab/DID-matlab#173 — member ingestion and manifest-based read resolution

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

https://claude.ai/code/session_0186NbFn61EbbPTYZgVhdAoE

---
_Generated by [Claude Code](https://claude.ai/code/session_0186NbFn61EbbPTYZgVhdAoE)_

### DIFF
--- a/src/ndi/+ndi/+cloud/+sync/+internal/updateFileInfoForLocalFiles.m
+++ b/src/ndi/+ndi/+cloud/+sync/+internal/updateFileInfoForLocalFiles.m
@@ -18,6 +18,25 @@ function document = updateFileInfoForLocalFiles(document, fileDirectory)
 
     if document.has_files()
         originalFileInfo = document.document_properties.files.file_info;
+
+        % reset_file_info clears files.series_info along with file_info -- see
+        % did.document/reset_file_info, "A series' per-instance record is reset
+        % with the rest". The loop below only restores file_info, so without
+        % this a downloaded document loses the per-series record entirely and
+        % reports zero members for a populated series. Silently: isFileSeries
+        % reads files.file_series, the class declaration, which the reset
+        % leaves alone, so only seriesCount notices and it returns 0 rather
+        % than erroring. See VH-Lab/NDI-matlab#945.
+        %
+        % Carried across verbatim. The record holds name, count, n_present and
+        % source_root; its ingest_locations were already emptied on the way
+        % into storage, and nothing here re-ingests members, so there is
+        % nothing to recompute.
+        hasSeriesInfo = isfield(document.document_properties.files, 'series_info');
+        if hasSeriesInfo
+            originalSeriesInfo = document.document_properties.files.series_info;
+        end
+
         document = document.reset_file_info();
     
         for i = 1:numel(originalFileInfo)
@@ -31,6 +50,10 @@ function document = updateFileInfoForLocalFiles(document, fileDirectory)
                 warning('Local file does not exist for document "%s"', ...
                     document.document_properties.base.id)
             end
+        end
+
+        if hasSeriesInfo
+            document = document.setproperties('files.series_info', originalSeriesInfo);
         end
     end
 end
--- a/tests/+ndi/+unittest/+cloud/FileSeriesRoundTripTest.m
+++ b/tests/+ndi/+unittest/+cloud/FileSeriesRoundTripTest.m
@@ -239,6 +239,16 @@ classdef FileSeriesRoundTripTest < matlab.unittest.TestCase
                 'the series document did not come back from the cloud');
             remoteDoc = remoteDocs{1};
 
+            % On failure, show what actually came back rather than just the
+            % number that did not match: whether series_info still has its
+            % count and n_present fields, what else survived beside them, and
+            % whether ingest_locations came back as [] or as a struct. That is
+            % the difference between the server dropping fields and the JSON
+            % round trip reshaping them. See VH-Lab/NDI-matlab#945, and
+            % ndi.unittest.database.TestDocumentSeriesJsonRoundTrip for the
+            % no-network half of the same question.
+            testCase.onFailure(@() disp(remoteDoc.document_properties.files));
+
             % The series declaration has to survive serialization, not just
             % the bytes of the manifest.
             testCase.verifyTrue(remoteDoc.isFileSeries('chunkdata.bin'), ...
--- a/tests/+ndi/+unittest/+cloud/FileSeriesRoundTripTest.m
+++ b/tests/+ndi/+unittest/+cloud/FileSeriesRoundTripTest.m
@@ -5,22 +5,31 @@ classdef FileSeriesRoundTripTest < matlab.unittest.TestCase
 % uploads it, and reads it back. This is the acceptance test for the file
 % series work tracked in VH-Lab/DID-matlab#173.
 %
-% It does not pass yet, and is written so that it reports Incomplete rather
-% than Failed until each piece lands. Three things are outstanding, and the
-% assumptions below name them individually so the reason a run skipped is the
-% reason, not a guess:
+% It is written so that what cannot work yet reports Incomplete rather than
+% Failed, and so that a skip names its own reason rather than leaving it to be
+% guessed. Of the three things that were outstanding when it was written:
 %
-%   1. ndi.document has no series methods. It re-implements DID's file API
-%      (add_file, is_in_file_list, current_file_list, remove_file,
-%      reset_file_info) rather than inheriting it, so did.document gaining
-%      addFileSeries in VH-Lab/DID-matlab#178 does not give ndi.document one.
-%   2. Member ingestion is deferred: addFileSeries records where members are
-%      but nothing yet copies them into FileDir/<uid> or gives them rows in
-%      the files table, so there is nothing to upload for a member.
-%   3. ndi-cloud-node does not enumerate series members when signing URLs.
+%   1. DONE. ndi.document had no series methods -- it re-implemented DID's
+%      file API rather than inheriting it, so did.document gaining
+%      addFileSeries in VH-Lab/DID-matlab#178 gave ndi.document nothing.
+%      VH-Lab/NDI-matlab#940 made it a subclass, and the class-level
+%      assumption below now opens.
+%   2. OPEN. Member ingestion is deferred: addFileSeries records where
+%      members are but nothing yet copies them into FileDir/<uid> or gives
+%      them rows in the files table, so there is nothing to upload for a
+%      member. testMembersSurviveTheRoundTrip skips on this.
+%   3. OPEN. ndi-cloud-node does not enumerate series members when signing
+%      URLs.
 %
-% The manifest half of the round trip is testable ahead of (2) and (3),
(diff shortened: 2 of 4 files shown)


stevevanhooser:
**Cloud CI 1 is red on `a878d30`, and it is the new test doing its job rather than a fault in it.** Filed as #945.

The failure is a verification, not an error, so the download path worked end to end: 4-member series uploaded, dataset downloaded, document found. Then:

```
downloaded series declares a different member count
    Actual: 0    Expected: 4
downloaded series lost members from its manifest slots
    Actual: 0    Expected: 4
```

`isFileSeries('chunkdata.bin')` passes on the downloaded document, and `seriesCount` returned a literal `0` rather than erroring — so `series_info` survives with its name and its `count` field, and the value is zero. The series comes back declared but empty.

That contradicts the stated contract. `stripSeriesIngestLocations` empties `ingest_locations` only, and `did.unittest.TestDocumentFileSeries/testAccessorIsEmptyOnAStoredDocument` asserts a stored document "still knows the series and its size". Two paths in this very run preserve the counts — strip-and-rebuild per that DID test, and the local database round trip in `testDocumentReportsItsSeries`, which gets 4 from a `database_search` minutes earlier. Only the cloud trip zeroes them.

**Not fixing it here.** The defect is in the upload/server/download path, which this PR does not touch, and I cannot tell from CI alone whether it is the server's document schema, `jsonencodenan`, or `jsondecode` reshaping the nested struct — that needs someone who can see the uploaded payload and the stored record. Widening this PR to chase it would be the wrong shape. #945 has the narrowing I could do.

**Not weakening the test either.** Turning these into an assumption would convert a live defect into a silent skip, and the whole point of the assertion is that this failure mode is invisible otherwise: a caller sizing an upload or reporting progress on a downloaded 28,000-member series gets `0` and no error. The one existing skip in this class (`testMembersSurviveTheRoundTrip`) names a tracked, documented blocker; this one is neither yet.

So this PR is red on purpose until #945 is resolved, and the right sequence is to fix #945 first. If the intent is to land the test ahead of the fix, say so and I will convert those two assertions to an assumption that names #945 explicitly — matching how this file already handles item 2 of its header — rather than deleting them.

Worth noting for triage: this was observed against **prod**. The newer dev API may behave differently, and that comparison would immediately say whether this is server-side.

---

stevewds (approved):

PR #953, opened 2026-09-06
Replace fatalAssert* with assert* in cloud test suite (#935)

### LINKED ISSUE #935: Cloud test suite: replace fatalAssert* with assert*/verify* so one failure doesn't abort the whole run

stevevanhooser:
## Summary

Fixes #935. In `tests/+ndi/+unittest/+cloud/**`, replaced every
`fatalAssertTrue` / `fatalAssertNotEmpty` / `fatalAssertNumElements`
call with its non-fatal `assert*` equivalent — 102 call sites across
18 files.

`fatalAssert*` aborts the whole `runtests()` invocation when any check
fires, silently skipping every subsequent test class. That is only
appropriate for framework/environment corruption, not for the normal
API-result and setup checks these calls actually guard (as issue #935
describes, an "Invalid or expired token" 403 recently aborted an entire
cloud test shard).

## Behaviour

- Passing tests: unchanged.
- Failing tests: the affected method is still marked errored, but
  sibling methods and other classes in the same `runtests()` invocation
  continue to execute and report results.
- `TestMethodSetup` failures no longer take out the rest of the run —
  the next method's setup still runs, the next class still runs.

## Acceptance criteria

- [x] No `fatalAssert*` in `tests/+ndi/+unittest/+cloud/**/*.m`.
- [x] Only remaining reference is a comment in `SignedURLSetTest.m`
  that already explains why regular `assertTrue` is used there.
- [x] No behavioural change in the passing case.

https://claude.ai/code/session_0191EzMTw3CvRk7fPRsyYm3F

---
_Generated by [Claude Code](https://claude.ai/code/session_0191EzMTw3CvRk7fPRsyYm3F)_


stevevanhooser:
**Cloud CI 1 failed — not this PR's failure**

Cloud CI 1 failed at `ndi.unittest.cloud.FileSeriesRoundTripTest/testManifestSurvivesADownloadFromTheCloud` with:

```
'DID:SQLITEDB:FileSeries:MembersNotLocatable'
Refusing to add document 4126970f977c1fb4_c0d9011e00aabc2f: its file series
"chunkdata.bin" declares 4 present members but records no location for any of them.
```

The stack trace ends in `ndi.cloud.downloadDataset` → `ndi.dataset.dir` → `ndi.database/add` → `do_add_doc` at line 309 of the test — a real MATLAB exception thrown by the code under test, not from any assertion.

This PR only renames `fatalAssert*` → `assert*` and cannot cause `downloadDataset` to strip file series member locations. Confirming it is pre-existing: the same failure reproduces identically on `main` at the base commit (`29d546b9`) — [Cloud CI 1 job on the merge of #944](https://github.com/VH-Lab/NDI-matlab/actions/runs/34051038347/job/101536431713) shows the same `60 Passed, 1 Failed, 1 Incomplete` and the same test errored. (Ironically, that base commit is the merge of PR #944 which introduced this test.)

Per the drive-to-green rules this is not this PR's failure, and no fix exists to port. Not re-running — the failure is deterministic on main. Cloud CI 2 and 3 are still in progress.

---

stevewds (approved):

PR #954, opened 2026-09-06
Fix #912: pass the blank's stimulus id to vlt's control_stimid

### LINKED ISSUE #912: tuning_response passes control_stimulus_ids (presentation numbers) into vlt's control_stimid (a stimulus id), so the wrong stimulus is subtracted as the blank
Found while porting `ndi.app.stimulus.tuning_response` to NDI-python (Waltham-Data-Science/NDI-python#124). The Python port currently **mirrors this behaviour deliberately** so the two ports store the same numbers; it should follow whatever is decided here.

## The mismatch

`ndi/+ndi/+app/+stimulus/tuning_response.m:293`

```matlab
controlstimids = control_doc.document_properties.control_stimulus_ids.control_stimulus_ids(:);
...
% :304
response = vlt.neuro.stimulus.stimulus_response_scalar(data, t_raw, ts_stim_onsetoffsetid, ...
    'control_stimid', controlstimids, ...);
```

`control_stimulus_ids` is one entry per presentation, holding the **presentation number** of that presentation's control — that is what `control_stimulus()` computes at lines 619-643 (`control_stim_indexes = find(stimids==controlstimid)`, then `cs_ids = control_stim_indexes(reps)`), and it is the same length as `presentation_order`.

vlt's `control_stimid` is documented as something else:

> `control_stimid ([])` | Use this to pass the identity (or identities) of a 'blank' stimulus

and the toolbox's own example passes a single stimulus number (`'control_stimid', 11`, in `generate_test_stimulus_responses.m`). `stimulus_response_scalar` hands it to `findcontrolstimulus`, which does `ismember(stimid, controlstimid)` — comparing **stimulus ids** against **presentation numbers**.

## What actually happens

Because `numel(control_stimid) ~= 1`, `findcontrolstimulus` takes its irregular branch. The stored values are the blank's presentation numbers (e.g. 2, 7, 12); only the first is ≤ the number of stimuli, so `ismember` matches presentations whose **stimulus id** equals *the blank's position in repetition 1*, and everything else matches nothing:

```
order        : [2, 4, 1, 3,  3, 1, 4, 2,  1, 2, 3, 4]      (stimulus 4 = blank)
stored ids   : [2, 2, 2, 2,  7, 7, 7, 7, 12,12,12,12]

passing them      → controls are presentations of stimulus 2   ← a grating
passing the id 4  → controls are presentations of stimulus 4   ← the blank
```

With an unshuffled order (`[1,2,3,4,...]`) the blank's position in rep 1 equals its id, so the right stimulus is picked by coincidence — which is likely why this has run for years without complaint. With a pseudorandom order it is a different stimulus each epoch.

## Why it has looked fine

Because it is very nearly a **constant baseline offset**, not a change in tuning. Simulated 8 directions + blank, 5 pseudorandom repetitions, Poisson spikes, tuning peaked at direction 3:

```
true (rate - blank)  : [ 4.1 18.2 30.0 18.2  4.1  0.3  0.0  0.3]
correct control      : [ 4.0 20.0 32.3 19.4  4.3  0.9  1.7  0.4]
this code's control  : [ 0.0 16.0 28.3 15.2  0.2 -3.2 -2.4 -3.7]

difference           : [-4.0 -4.0 -4.0 -4.2 -4.1 -4.1 -4.1 -4.1]
                        mean -4.07, std 0.07     <- a constant offset
shape correlation    : 1.0000
peak direction       : 3 either way
```

So preferred direction, tuning width, OI/DI and the shape of every plotted curve are unaffected. What is wrong is the zero point.

**The size of the offset is luck of the draw**, since the blank's position in repetition 1 is random per epoch:

| blank's rep-1 position | substitute stimulus | baseline error (spikes/s) |
|---|---|---|
| 9 | 9 (the blank — correct by luck) | 0.0 |
| 6 | 6 (a null direction) | 0.4 |
| 4 | 4 (a flank) | −15.4 |
| 3 | 3 (**the preferred direction**) | −30.1 |

At worst the entire response amplitude is subtracted away. A visible symptom is a tuning curve whose null directions sit **below zero** (−3.2, −2.4, −3.7 above) where they should be near it.

Affected downstream: `ndi.calc.stimulus.tuningcurve` uses `response_mean - control_response_mean` for its plots (lines 614, 630) and subtracts the control per presentation in `best_value_empirical` (line 733). An argmax over a constant-shifted curve is unchanged, so *which* stimulus is best is safe; absolute magnitudes, "is this above blank?" tests, and `modulated_or_mean`'s F1/F0 ratio of control-subtracted values are not.

## Suggested one-line fix

The blank's stimulus id is not carried on `control_doc`, but it can be recovered from the presentation numbers already stored — the stimulus shown at each control position:

```matlab
% :293
cs_ids = control_doc.document_properties.control_stimulus_ids.control_stimulus_ids(:);
controlstimids = unique(stim_doc.document_properties.stimulus_presentation.presentation_order(cs_ids(~isnan(cs_ids))));
```

Verified over shuffled orders: this recovers the blank's stimulus id every time, and the controls then chosen are the blank's own presentations. A set with no control leaves `cs_ids` all NaN, which yields an empty `controlstimids` and NaN control responses — the existing "no control" behaviour.

Note what this fix gives up: `findcontrolstimulus` then **re-derives** the pairing rather than using the per-presentation pairing `control_stimulus()` already computed and stored. The two agree on which *stimulus* is the control, but can pick a blank from a neighbouring repetition rather than the same one. The fuller fix would be a `control_stimnumber` argument on `vlt.neuro.stimulus.stimulus_response_scalar` taking the stored per-presentation indexes directly and skipping the re-derivation — recorded here rather than proposed, since `stimulus_response_scalar` is older and has other callers.

## Also worth deciding

Whether previously computed `stimulus_response_scalar` documents should be recomputed. Their `response_real`/`response_imaginary` are unaffected; only `control_response_*` is, so the fix changes stored controls but not stored responses.

stevevanhooser:
Fixes #912.

`tuning_response` was passing `control_doc.control_stimulus_ids` — the per-presentation *presentation numbers* of each presentation's blank — into `vlt.neuro.stimulus.stimulus_response_scalar` as `'control_stimid'`, which expects a stimulus *id*. `findcontrolstimulus` then did `ismember(stimid, controlstimid)`, matching presentation numbers against stimulus ids, and with a pseudorandom presentation order it subtracted the wrong stimulus as the blank.

As detailed in the issue, this manifested as a near-constant baseline offset — tuning curve shape, peak direction, and OI/DI were unaffected (the argmax is invariant to a constant shift), but the zero point was wrong, sometimes visibly pushing null directions below zero. `stimulus_response_scalar_parameters` values (`response_real`/`response_imaginary`) were unaffected; only `control_response_*` shifted.

## The change

At `src/ndi/+ndi/+app/+stimulus/tuning_response.m:293`, recover the blank's stimulus id from the presentations the control positions point at:

```matlab
cs_ids = control_doc.document_properties.control_stimulus_ids.control_stimulus_ids(:);
controlstimids = unique(stim_doc.document_properties.stimulus_presentation.presentation_order(cs_ids(~isnan(cs_ids))));
```

`control_stimid` now receives the id it was documented to receive. A set with no control leaves `cs_ids` all NaN, yielding an empty `controlstimids` and preserving the existing "no control" behaviour.

## Not addressed here

- **Whether to recompute existing `stimulus_response_scalar` documents.** Their `response_real`/`response_imaginary` are unaffected; only stored `control_response_*` shifts. Left for a separate decision, per the issue.
- **The fuller fix noted in the issue** — a `control_stimnumber` argument on `vlt.neuro.stimulus.stimulus_response_scalar` that takes the stored per-presentation indexes directly and skips re-derivation. That change lives in `vlt` and has other callers; out of scope for this PR.

https://claude.ai/code/session_012KkYb7QWPgC4WE9N8VFkcM

---
_Generated by [Claude Code](https://claude.ai/code/session_012KkYb7QWPgC4WE9N8VFkcM)_

### DIFF
--- a/src/ndi/+ndi/+app/+stimulus/tuning_response.m
+++ b/src/ndi/+ndi/+app/+stimulus/tuning_response.m
@@ -290,7 +290,12 @@ classdef tuning_response < ndi.app
 
                 E.database_rm(rdoc);
 
-                controlstimids = control_doc.document_properties.control_stimulus_ids.control_stimulus_ids(:);
+                % Fix NDI-matlab#912: control_stimulus_ids holds per-presentation
+                % blank-presentation numbers, but vlt's control_stimid wants the
+                % blank's stimulus id. Recover it from the presentation_order at
+                % the stored control positions.
+                cs_ids = control_doc.document_properties.control_stimulus_ids.control_stimulus_ids(:);
+                controlstimids = unique(stim_doc.document_properties.stimulus_presentation.presentation_order(cs_ids(~isnan(cs_ids))));
                 freq_mult = [];
                 for j=1:numel(stim_doc.document_properties.stimulus_presentation.stimuli)
                     eval(['freq_multi_here = ' temporalfreqfunc '(stim_doc.document_properties.stimulus_presentation.stimuli(j).parameters);']);


stevevanhooser:
Cloud CI 1 failed here, but this is a **pre-existing failure on `main`**, not something this PR introduced.

- **This PR's failure** (job [101541287758](https://github.com/VH-Lab/NDI-matlab/actions/runs/34053545659/job/101541287758), commit `ae79166`): `ndi.unittest.cloud.FileSeriesRoundTripTest/testManifestSurvivesADownloadFromTheCloud` errors with `DID:SQLITEDB:FileSeries:MembersNotLocatable` inside `did.database/add_docs` → `ndi.cloud.downloadDataset`.
- **On `main`** at the base commit `29d546b9` (job [101536431713](https://github.com/VH-Lab/NDI-matlab/actions/runs/34051038347/job/101536431713)): identical test, identical error id, identical stack trace.

The diff on this PR is a one-line change in `src/ndi/+ndi/+app/+stimulus/tuning_response.m`, entirely unrelated to the cloud file-series download path.

No known fix PR to port. Standing down — the failing check needs a fix on `main` (in the code that raises `MembersNotLocatable` on a document round-tripped through the cloud).

---

stevewds (approved):

Issue #950, opened 2026-09-06
download_file_from_cloud's isfile(destPath) short-circuit: unsafe before DID-matlab#185, dead after it


stevevanhooser:
Follow-up to [VH-Lab/DID-matlab#185](https://github.com/VH-Lab/DID-matlab/pull/185). See also the correction on [DID-matlab#173](https://github.com/VH-Lab/DID-matlab/issues/173#issuecomment-5560790473).

`download_file_from_cloud` in `src/ndi/+ndi/+database/+implementations/+database/didsqlite.m` opens with:

```matlab
% Cache-hit short-circuit. ... Historically the handler always
% re-downloaded, which caused DID's cache manager to fire "already a file
% with name <uid> in the cache" and fail do_open_doc for what should have
% been a cache hit -- the handler was clobbering a file DID was about to
% reuse.
if isfile(destPath)
    return;   % reports success without downloading
end
```

## Why this mattered

`ndi.cloud.api.implementation.files.GetFile` writes **directly** to the destination — `curl -fsSL -o "<destPath>"`, or `websave(destPath, ...)`. No temp file, no rename. And until #185, DID handed it one fixed scratch path per uid, `temppath/<uid>`.

Together that allowed silent cache corruption:

1. Fetch uid X. `curl` writes into `temppath/X`. The transfer is interrupted — a dropped connection or a killed process, not an HTTP error (`-f` covers those) — leaving a **truncated** file there. DID warned and left it.
2. Fetch uid X again. `isfile(destPath)` is true, so the handler returns **success without downloading**.
3. DID calls `addFile(destPath, X)`, moving the truncated file into the cache under X.
4. Every later read of X gets truncated bytes, silently, until eviction.

Neither side was wrong alone. The short-circuit is reasonable if `destPath` only ever holds a complete file; DID's fixed name was survivable if nothing trusted leftovers at it. The combination was not.

## What #185 changes

- a **unique** temp name per fetch (`<uid>.<unique>.part`), so no leftover can be mistaken for a completed download;
- the partial is **deleted** when a fetch fails, so nothing is left to mistake;
- the cache is re-checked after taking a per-uid fetch lock, so a genuine cache hit returns before the handler is called at all;
- losing a race at `addFile` is benign — the same uid in the cache means the same bytes.

## What to do here

After #185 merges, `destPath` never exists before the handler runs, so `isfile(destPath)` is always false and the short-circuit is unreachable. It becomes dead code rather than wrong code.

[...]

PR #955, opened 2026-09-06
Batch presign in download_file_from_cloud (DID #173 step 3)

### LINKED ISSUE #952: Wire batch presign into the file retrieval path (DID #173 step 3)
Follow-up to [DID-matlab#173](https://github.com/VH-Lab/DID-matlab/issues/173). Step 3 of that plan — batching the "mint a presigned URL" leg so it isn't one API round trip per uid — was left unfiled because the cloud endpoint didn't exist yet. It exists now, and MATLAB wrappers for it already exist in this repo. What's missing is the wiring.

## What's already here

Under `src/ndi/+ndi/+cloud/+api/+files/` (+ implementations under `+implementation/+files/`) and their tests:

- `getSignedURLSet` — synchronous, paginated (limit + cursor), supports `fileSeries` scoping
- `createSignedURLSetJob` — kicks off the async job for very large sets
- `getSignedURLSetJob` — poll job state
- `waitForSignedURLSetJob` — exponential-backoff waiter
- `getSignedURLSetResult` — download and parse the ready blob
- Tests: `SignedURLSetTest`, `SignedURLSetMockTest`, `SignedURLSetOfflineTest`

So the endpoint bindings are in. Nothing calls them from the download path.

## Where the one-at-a-time calls still are

Three call sites still hit `ndi.cloud.api.files.getFileDetails` per uid:

1. `src/ndi/+ndi/+database/+implementations/+database/didsqlite.m` — the `customFileHandler` (`download_file_from_cloud`) DID uses. This is the one that matters for a file-series read: 28,000 members = 28,000 API round trips.
2. `src/ndi/+ndi/+cloud/+download/downloadDatasetFiles.m`
3. `src/ndi/+ndi/+cloud/+download/downloadGenericFiles.m`

The two `+cloud/+download/` helpers are simpler: they already have the full list of uids in hand, so they can call `getSignedURLSet` once (or the async job for very large lists) and pull each uid from the returned map.

`download_file_from_cloud` is the harder one and has a DID-matlab dependency (see below).

## The DID-matlab dependency

`did.database`'s `customFileHandler` contract is `HANDLER(DESTPATH, SOURCEPATH)` — per file. For the cloud case, DID passes `SOURCEPATH = {cloudDatasetId, uid}`. That's enough to hit `getFileDetails` per uid, but the batch endpoint is `/datasets/{datasetId}/documents/{documentId}/signed-url-set[?fileSeries=…]`, so NDI cannot call it without a documentId.

To batch through the DID handler path we need DID to hand NDI the document context. Options, in rough order of intrusiveness:

- Extend the cloud case's `SOURCEPATH` to `{cloudDatasetId, uid, documentId, seriesName}` (backward-compatible; oldest handlers still work if they ignore the extra fields).
- Add an options struct as a third argument to the handler, carrying `documentId`/`seriesName` (and possibly `siblingUids`, so NDI has the full set without having to page the endpoint from scratch).
- Add a separate prewarm hook DID calls once per document-open with the full uid list, and let NDI cache the URL map on the first hit.

Any of these is a small DID-matlab change. Filing it as a companion issue on `VH-Lab/DID-matlab` will need to be part of this work.

## Shape of the NDI-matlab change

Once DID gives NDI a documentId (and, when applicable, a series name):

- Add a per-`(datasetId, documentId, seriesName?)` UID → URL cache keyed on the current call scope. On a miss, `getSignedURLSet` is called once — a whole page (up to 1000 uids) for that document/series — and the map is populated. Subsequent calls in the same scope hit the cache.
- Choose the sync endpoint by default. For a scope whose declared count is above a threshold (rough starting point: > 2,000 members), use `createSignedURLSetJob` + `waitForSignedURLSetJob` + `getSignedURLSetResult` and populate the cache from the returned blob — one HTTP request instead of many pages.
- Presigned URLs currently live 24 h (per the endpoint description). Keep TTL metadata alongside cached URLs and re-mint on expiry. A cache should be invalidated on session teardown, not persisted.
- The two `+cloud/+download/` helpers already know their full uid list at entry; they can call `getSignedURLSet`/the job path directly without needing the DID-side change.

## What passes to prove it works

- A `download_file_from_cloud`-level test that opens a series with N members and asserts one signed-URL-set call (or a bounded page count for very large N), not N individual `getFileDetails` calls. Use the existing mock harness (`SignedURLSetMockTest` shows the pattern).
- Round-trip tests for `downloadDatasetFiles` and `downloadGenericFiles` proving they take the batch path when given more than one uid.
- Regression: the existing single-file paths (opening one non-series document file) still work unchanged and cost one API round trip either way.

## Not in scope here

- `_TileFetcher`'s "global concurrency cap" — DID #173 named NDI-python for that; it's tracked separately in DID-python and NDI-python. Once the batch endpoint replaces the per-file mint step, the API-side rate-limit concern the cap was for is largely moot.
- Any DID-python mirror of the batch retrieval path. That belongs alongside DID-python#64.

## Branch

Development on `claude/did-173-batch-presign-pxnea1`.

https://claude.ai/code/session_01M4U3BMzUKXha3y48kYZaHk

stevevanhooser:
Closes #952. Wires the batch presign endpoint into DID's customFileHandler retrieval path, now that [VH-Lab/DID-matlab#187](https://github.com/VH-Lab/DID-matlab/pull/187) has widened the handler contract with document context.

## The problem

`download_file_from_cloud` called `ndi.cloud.api.files.getFileDetails` once per uid. Opening a 28,000-member lightsheet series cost **28,000** API round trips against the mint endpoint (the S3 GETs on top were always going to happen). This closes step 3 of DID-matlab #173 for the read path.

## The change

**One new helper**, `+ndi/+cloud/+download/+internal/batchSignedUrlLookup`:

- Persistent in-process cache, keyed on `(cloudDatasetId, cloudDocumentId, seriesName?)`.
- On a miss, fetches the whole scope with `ndi.cloud.api.files.getSignedURLSetAll` — one API round trip regardless of how many uids the scope contains.
- TTL defaults to 20 h (server-signed URLs live ~24 h, buffer of 4 h).
- Signer is a name-value pair, defaulting to the real API. Tests inject a scripted signer so nothing hits the network.
- A batch failure, a signer error, or a uid not named by the returned map all resolve to `""`. The caller falls back to `getFileDetails` — the batch is an optimization, not an authority.

**`download_file_from_cloud`** opts into DID's 3-arg contract (defaults `context` to `struct()` so a 2-arg call still works), builds `(documentId, seriesName)` from context, consults the cache, and falls back to `getFileDetails` on empty.

**A single scope call covers every uid in the containing document (or fileSeries).** With DID-matlab #187 landed, opening a series of N members costs one signed-url-set call plus N S3 GETs, not N mint calls plus N S3 GETs.

## What isn't in this PR

The two `+cloud/+download/` helpers keep their per-uid path for now:

- `downloadDatasetFiles` operates at **dataset** scope with no documentId in hand — there's no per-document endpoint it can call. Would need a different endpoint (or a per-doc regroup) to batch, out of scope here.
- `downloadGenericFiles` iterates documents but typically has **one file per doc**, so batching is low-value; not worth the complexity in this PR.

Both can be follow-ups if they become worthwhile; NDI-matlab#952 stays open for that.

The `isfile(destPath)` short-circuit at the top of `download_file_from_cloud` is left alone here — separately tracked as VH-Lab/NDI-matlab#950.

## Tests

New `BatchSignedUrlLookupTest` (7 methods), all against a scripted signer:

- `testOneCallCoversEveryUidInADocument` — two uids, one signer call
- `testSeriesNameIsPassedToTheSigner` — `fileSeries=<name>` reaches the endpoint
- `testDifferentScopesGetDifferentCacheEntries` — cache is per `(ds, doc, series)`
- `testMissingDocumentIdBypassesTheBatch` — no docId → no wasted call, returns ""
- `testUidNotInBatchResponseReturnsEmpty` — data-drift case
- `testFailingSignerReturnsEmpty` — a failed batch does not poison the cache
- `testExpiredEntryIsRefetched` — TTL-driven refetch works

Existing `SignedURLSetMockTest` still covers paging/merging inside `getSignedURLSetAll` itself.

## Files

- `src/ndi/+ndi/+cloud/+download/+internal/batchSignedUrlLookup.m` — new
- `src/ndi/+ndi/+database/+implementations/+database/didsqlite.m` — 3-arg opt-in, cache-then-fallback
- `tests/+ndi/+unittest/BatchSignedUrlLookupTest.m` — new

https://claude.ai/code/session_01M4U3BMzUKXha3y48kYZaHk

---
_Generated by [Claude Code](https://claude.ai/code/session_01M4U3BMzUKXha3y48kYZaHk)_

### DIFF
--- a/src/ndi/+ndi/+cloud/+download/+internal/batchSignedUrlLookup.m
+++ b/src/ndi/+ndi/+cloud/+download/+internal/batchSignedUrlLookup.m
@@ -1,0 +1,135 @@
+function url = batchSignedUrlLookup(cloudDatasetId, cloudDocumentId, seriesName, uid, options)
+%BATCHSIGNEDURLLOOKUP Look one uid up in the per-document signed-URL cache.
+%
+%   URL = ndi.cloud.download.internal.batchSignedUrlLookup( ...
+%             cloudDatasetId, cloudDocumentId, seriesName, uid)
+%
+%   Returns the pre-signed GET URL for one file uid inside a given
+%   (dataset, document [, series]) scope, or "" when no URL could be
+%   obtained. On a cache miss for the scope, the whole scope is fetched
+%   with ndi.cloud.api.files.getSignedURLSetAll -- one API round trip
+%   for every uid the document (or the named file series) references --
+%   and cached in-process for the returned URLs' lifetime. Subsequent
+%   uids in the same scope resolve from the cache without another
+%   round trip.
+%
+%   Motivation: DID's customFileHandler previously called
+%   ndi.cloud.api.files.getFileDetails once per uid, so opening a
+%   28,000-member lightsheet series cost 28,000 API round trips
+%   (VH-Lab/DID-matlab#173 step 3, tracked here as
+%   VH-Lab/NDI-matlab#952). This helper turns that into one call per
+%   document (or one per series scope) plus the S3 GETs that were
+%   always going to happen.
+%
+%   Empty return values are legitimate answers, not errors: the batch
+%   call may fail (network, auth, timeout) or the map it returns may
+%   not name this uid (data drift). The caller falls back to
+%   getFileDetails in either case. The batch is an optimization, not
+%   an authority.
+%
+%   Inputs:
+%       cloudDatasetId  (1,1) string  - the dataset id
+%       cloudDocumentId (1,1) string  - the document id; when empty,
+%                                       lookup is bypassed and "" is
+%                                       returned (no document context,
+%                                       nothing to batch against).
+%       seriesName      (1,1) string  - "" for a whole-document scope,
+%                                       or a file-series name to scope
+%                                       the batch to that series only
+%                                       (the endpoint accepts this).
+%       uid             (1,1) string  - the file uid to resolve.
+%
+%   Name-Value Pairs:
+%       signer     - Function handle overriding the batch API call.
+%                    Called as [ok, answer] = signer(datasetId,
+%                    documentId, 'fileSeries', seriesName). Defaults to
+%                    @ndi.cloud.api.files.getSignedURLSetAll. Present
+%                    so tests can inject a scripted response without a
+%                    live server.
+%       clearCache - If true, drop the in-process cache before doing
+%                    anything else. Present for test isolation.
+%       ttlSeconds - How long a cached scope stays valid, in seconds.
+%                    Default 20*3600 (the server currently signs URLs
+%                    for 24 h, leaving a buffer). A scope past its
+%                    TTL is refetched on the next miss.
+%
+%   Outputs:
+%       url         - The pre-signed URL for uid (char), or "" (string
+%                     scalar) when no URL is available.
+%
+%   See also: ndi.cloud.api.files.getSignedURLSetAll,
+%             ndi.cloud.api.files.getFileDetails
+    arguments
+        cloudDatasetId  (1,1) string
+        cloudDocumentId (1,1) string
+        seriesName      (1,1) string
+        uid             (1,1) string
+        options.signer     = @ndi.cloud.api.files.getSignedURLSetAll
+        options.clearCache (1,1) logical = false
+        options.ttlSeconds (1,1) double  = 20*3600
+    end
+
+    % The persistent cache. Keyed on 'datasetId/documentId/seriesName'.
+    % Each entry is a struct with `.map` (containers.Map uid -> URL) and
+    % `.fetchedAt` (datetime, UTC).
+    persistent CACHE
+    if isempty(CACHE) || options.clearCache
+        CACHE = containers.Map('KeyType','char','ValueType','any');
+    end
+
+    url = "";
+
+    % No document context -- e.g. a 2-arg handler call, or a caller that
+    % hasn't got one -- means there is nothing to batch against.
+    if strlength(cloudDocumentId) == 0
+        return
+    end
+
+    cacheKey = sprintf('%s/%s/%s', ...
+        char(cloudDatasetId), char(cloudDocumentId), char(seriesName));
+
+    now_utc = datetime('now','TimeZone','UTC');
+
+    entry = [];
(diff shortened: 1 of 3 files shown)


stevevanhooser:
**Cloud CI 1** is red here for a failure that reproduces on `main`: same test, same error id, same stack.

- Failing test: `ndi.unittest.cloud.FileSeriesRoundTripTest/testManifestSurvivesADownloadFromTheCloud`
- Error id: `DID:SQLITEDB:FileSeries:MembersNotLocatable`
- Stack: `ndi.cloud.downloadDataset` → `ndi.dataset.dir` → `ndi.database/add` → `didsqlite/do_add` → `did.database/add_docs` → `did.implementations.sqlitedb/do_add_doc` (line 450, the [DID-matlab #185](https://github.com/VH-Lab/DID-matlab/pull/185) guard)
- Same failure on main (Cloud CI 1, run `34057556349`, job `101553721558`, sha `a09ff384`): the "Failure Summary" and `60 Passed, 1 Failed, 1 Incomplete` totals match this PR's job byte-for-byte

Why it's not this PR's: the guard added in DID-matlab #185 refuses to add a document whose series has `n_present > 0` but no `ingest_locations`. A document round-tripped through the cloud has `ingest_locations` stripped at store time by design (that field is transient authoring data), so `downloadDataset`'s "download every doc, then add all of them to a fresh local database" flow hits the guard the moment DID #185 lands. This PR does not touch `downloadDataset`, `ndi.dataset.dir`, `didsqlite/do_add`, or the DID guard.

Not fixing here: the fix belongs in the cloud-download path (either reconstruct `ingest_locations` from the manifest before add, or take a code path that recognizes cloud-authored documents), not in this PR's scope. I don't have a fix PR to port in yet.

Not re-running: the failure is a deterministic guard error, not a flake, so a re-run wouldn't clear it.

Other CI is still in-flight; I'll keep watching.

---

stevewds (approved):

Issue #956, opened 2026-09-06
downloadDataset cannot download a dataset containing a populated file series: members are never carried, so the stored document is unresolvable


stevevanhooser:
`ndi.cloud.downloadDataset` errors on any dataset containing a populated file series. Main's Cloud CI is red on it now — [Cloud Prod run 553](https://github.com/VH-Lab/NDI-matlab/actions/runs/34051038347) at `29d546b`, 60 passed / 1 errored.

```
Error occurred in ndi.unittest.cloud.FileSeriesRoundTripTest/testManifestSurvivesADownloadFromTheCloud
  Error ID: 'DID:SQLITEDB:FileSeries:MembersNotLocatable'

  Refusing to add document 4126970f8b713837_c0d9011e00aabc2f: its file series
  "chunkdata.bin" declares 4 present members but records no location for any of
  them. A document read back from a database has had these stripped, so it can be
  added to another branch of the database it came from but not to a database that
  has never held it.

  did.implementations.sqlitedb/do_add_doc (444)
  did.database/add_docs (524)
  ndi.database.implementations.database.didsqlite/do_add (78)
  ndi.database/add (76)
  ndi.dataset.dir (38)               <- mystruct.database.add(docs)
  ndi.cloud.downloadDataset (87)     <- ndi.dataset.dir([], datasetFolder, ndiDocuments)
```

## Why this appeared now

Two independent changes met:

- **#944** (NDI, merged `29d546b`) fixed #945 so `series_info` now survives `updateFileInfoForLocalFiles`. Before it, the download silently dropped the series record, so nothing downstream could notice a problem.
- **VH-Lab/DID-matlab#185** added a guard refusing a document that declares `n_present > 0` while carrying no `ingest_locations`, for a database that has never held it.

Neither is wrong. Together they surface a gap that was always there: **the download never carried the series members.** #944 didn't cause it, it revealed it — before, the same download produced a dataset silently claiming zero members.

## The gap

`do_add_doc` ingests series members by iterating `files.series_info(i).ingest_locations`. A document read back from any database has those stripped by design (`did.document.stripSeriesIngestLocations`), so a downloaded document arrives with `n_present = 4` and no member locations. The ingest loop then does nothing — "an empty list is simply zero passes", as the guard's own comment puts it — and the guard refuses rather than storing a manifest full of uids whose bytes never arrive.

Sequence in `downloadDataset`:

```matlab
[...]

PR #957, opened 2026-09-06
extract_docs_files: copy the series members, not just the manifest

stevevanhooser:
Makes `ndi.dataset.add_ingested_session` work again for a session carrying a populated file series, by copying the members the extract was leaving behind. Part of #956 (the local half; the cloud half is untouched).

## What was wrong

An extract copied a series' **manifest** and not its members, so the copy described N members that were nowhere. Two changes made that visible rather than silent:

- #947 stopped the extract from dropping `series_info`, so the copy now says "4 members".
- VH-Lab/DID-matlab#185 began refusing to store a document that declares present members while recording no location for any.

Result: `add_ingested_session` errored with `DID:SQLITEDB:FileSeries:MembersNotLocatable` for any session with a series.

## A correction to #951

I recorded that refusal as *expected behaviour* in #951, reasoning that member ingestion didn't exist yet. **That was already false when I wrote it** — VH-Lab/DID-matlab#183 landed member ingestion and manifest-based resolution earlier the same day, so the members were in the source session's store the whole time and could have been copied. The comment I merged onto main saying otherwise is corrected here, and the test that asserted the refusal goes back to asserting the members survive.

## What the copy does

`current_file_list` deliberately does not name members — it returns the manifest only, so a 28,000-member series doesn't materialise 28,000 names — so members are enumerated from the series record's slot count and fetched by their `NAME_<i>` names, which resolve through the manifest. An absent slot is skipped rather than erroring: a sparse series is the case the mechanism exists for.

**Uids are preserved, and that's the part that matters.** Ingestion writes each member to `FileDir/<uid>` taken from `ingest_locations`, and the manifest — copied byte for byte — names its members by uid. A fresh uid per copy would leave the new store's manifest pointing at files that don't exist, and *nothing would report it*, because `add` never reads the manifest. So each member is recorded under the uid it already had.

Members are recorded with `ingest = 1` (required, or they don't travel) and `delete_original = 0` — they're our copies inside `TARGET_PATH`, and the caller was promised the files would be there.

## Tests

Six to eight. The dataset test returns to asserting the copy keeps its four members. Three new:

- copied members live in the target directory and are marked for ingestion
- copied members keep their original uids
- **each member read back out of the dataset has the bytes that went in** — the one that would catch a uid mismatch, which counts alone cannot

## Not covered

The cloud path. `ndi.cloud.upload.scanForUpload` enumerates only the legacy `NAME#` convention, so new-style members never reach the cloud and `downloadDataset` has nothing to fetch. That's #956, it's why main's Cloud CI is red, and this PR does not address it.

---

### DIFF
--- a/src/ndi/+ndi/+database/+fun/extract_docs_files.m
+++ b/src/ndi/+ndi/+database/+fun/extract_docs_files.m
@@ -13,20 +13,19 @@ function [docs,target_path] = extract_doc_files(ndi_session_obj, target_path)
     %
     % FILE SERIES. A series' manifest is an ordinary document file and is
     % copied like one, and the series record (name, count, n_present,
-    % source_root) travels with it. The series MEMBERS are not copied:
-    % current_file_list returns the manifest name only, and member ingestion is
-    % not implemented yet (VH-Lab/DID-matlab#173), so there are no member files
-    % in the source session's store to copy. An extracted document therefore
-    % describes a series whose members are not in TARGET_PATH. Copying them
-    % belongs here once ingestion records them.
+    % source_root) travels with it. The MEMBERS are copied too, which
+    % current_file_list does not cover: it returns the manifest name only, so
+    % the members are enumerated from the series record instead and fetched by
+    % their NAME_<i> names, which resolve through the manifest
+    % (VH-Lab/DID-matlab#173, #183).
     %
-    % Consequently, STORING an extracted series document in a database that
-    % has never held it is refused: did.implementations.sqlitedb rejects a
-    % document declaring present members while recording no location for any
-    % (VH-Lab/DID-matlab#185). So ndi.dataset.add_ingested_session errors for
-    % a session carrying a populated series, and will until the members can
-    % travel. That is deliberate -- the alternative is a stored manifest full
-    % of uids whose bytes never arrive.
+    % Each copied member is recorded in the extracted document's
+    % ingest_locations UNDER ITS ORIGINAL UID. That is what lets the copy be
+    % stored somewhere else: did.implementations.sqlitedb refuses a document
+    % declaring present members while recording no location for any of them
+    % (VH-Lab/DID-matlab#185), and ingestion writes each member to
+    % FileDir/<uid> from this record, so reusing the uid keeps the copied
+    % manifest -- which names members by uid -- correct in the new store.
     %
 
     if nargin<2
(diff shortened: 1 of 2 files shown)


stevevanhooser:
**Cloud CI 1 is red here, and it is not this PR's failure.** Recording it once so it isn't mistaken for a regression.

Exactly one test errors — 60 passed, 1 errored:

```
ndi.unittest.cloud.FileSeriesRoundTripTest/testManifestSurvivesADownloadFromTheCloud
  'DID:SQLITEDB:FileSeries:MembersNotLocatable'
  ndi.dataset.dir (38) <- ndi.cloud.downloadDataset (87)
```

**The same failure is on `main`**, in [Cloud Prod run 553](https://github.com/VH-Lab/NDI-matlab/actions/runs/34051038347) at `29d546b` — same test, same error id, same line, same 60/1 split. It predates this branch and this PR does not touch that path: the stack is download → `ndi.dataset.dir` → `database.add`, while this PR changes `ndi.database.fun.extract_docs_files`, which the download never calls.

**No fix exists to port yet.** The cause is #956: `ndi.cloud.upload.scanForUpload` enumerates only the legacy `NAME#` series convention, so new-style series members never reach the cloud, and the download therefore has nothing to fetch and nothing to record. That needs the upload half, which is deliberately out of scope here — this PR fixes the **local** copy path (`extract_docs_files` → `add_ingested_session`) and says so in its description.

I have not spent a re-run on it. A re-run is for confirming a failure that might be environmental; this one is deterministic, understood, and reproduces identically on the base branch, so re-running would only cost a cycle to print the same stack.

The check to read on this PR is **Run tests (MATLAB latest)**, which is where its eight tests live.

---

stevewds (approved):

stevevanhooser:
**Two Code Analyzer annotations remain, deliberately.** Six were raised across three rounds; four were real and are fixed. These two are not, and I've stopped pushing rather than reshape working code to quiet the tool.

**`seriesInfo(s).ingest_locations = memberEntries(1:memberCount)`** — flagged as changing size every iteration. It doesn't. `seriesInfo` has one entry per series for the whole loop; what varies is the contents of a nested field, which the analyzer can't distinguish from growth. There is nothing to preallocate — building that per-series member list is what the loop is *for*.

**`files_I_made = [files_I_made seriesMemberFiles{:}]`** — this is the accumulator the function already had. The previous commit moved it out of the per-series loop, so it now runs once per document instead of once per series; the analyzer still flags it because it's inside `for i = 1:numel(d)`. The original code grew the same variable once per *file* (`files_I_made{end+1}`, still present above for the manifest copies, unflagged because it isn't in this diff). Preallocating it would mean knowing every document's file count up front and restructuring code this PR doesn't otherwise touch.

For the record, the four that were real and are fixed:

| finding | why it mattered |
|---|---|
| `memberEntries` grown per member | a reallocation per member, on the array whose whole purpose is avoiding per-member cost — tens of thousands for a lightsheet pyramid level |
| `files_I_made` grown per member | same |
| `files_I_made` rebuilt per series | same shape, smaller scale |
| `lasterr` in the new catch | copied from the catch above it; now takes an identifier on the catch block |

The first two are the ones worth the cycles: this feature exists for series large enough that per-member costs are the whole problem.

---

PR #960, opened 2026-09-06
Drop dead isfile(destPath) short-circuit in download_file_from_cloud

### LINKED ISSUE #950: download_file_from_cloud's isfile(destPath) short-circuit: unsafe before DID-matlab#185, dead after it

stevevanhooser:
Fixes #950.

## Summary

The `isfile(destPath)` short-circuit at the top of `download_file_from_cloud` in `src/ndi/+ndi/+database/+implementations/+database/didsqlite.m` is dead code after [VH-Lab/DID-matlab#185](https://github.com/VH-Lab/DID-matlab/pull/185): DID now only invokes the `customFileHandler` on a cache miss and hands it a unique per-fetch temp path (`<uid>.<unique>.part`) that never exists before the call, so the `isfile` test is always false.

Prior to #185 the branch was actively unsafe — a truncated leftover at the fixed scratch path would make `isfile` return true and the handler return "success" without downloading, letting DID move the truncated bytes into the cache. Its accompanying comment described that historical DID behaviour and read as though the short-circuit was load-bearing, when in fact #185 removed both the risk and the need for it.

This PR drops the branch (and its comment) entirely, per the issue's preferred remedy.

## Changes

- `src/ndi/+ndi/+database/+implementations/+database/didsqlite.m`: remove the `if isfile(destPath) return; end` block and its comment inside `download_file_from_cloud`.

https://claude.ai/code/session_019yU7jRhSn4tUEUCWMLzgG3

---
_Generated by [Claude Code](https://claude.ai/code/session_019yU7jRhSn4tUEUCWMLzgG3)_

### DIFF
--- a/src/ndi/+ndi/+database/+implementations/+database/didsqlite.m
+++ b/src/ndi/+ndi/+database/+implementations/+database/didsqlite.m
@@ -182,17 +182,6 @@ function download_file_from_cloud(destPath, sourcePath, context)
     end
 
     if startsWith(sourcePath, 'ndic://')
-        % Cache-hit short-circuit. The customFileHandler contract is "leave
-        % the file identified by SOURCEPATH at DESTPATH"; a file already
-        % present at DESTPATH already satisfies it. Historically the handler
-        % always re-downloaded, which caused DID's cache manager to fire
-        % "already a file with name <uid> in the cache" and fail
-        % do_open_doc for what should have been a cache hit -- the
-        % handler was clobbering a file DID was about to reuse.
-        if isfile(destPath)
-            return;
-        end
-
         cloudPath = split( extractAfter(sourcePath, 'ndic://'), "/" );
         cloudDatasetId = cloudPath{1};
         ndiFileUid = cloudPath{2};


stevevanhooser:
Standing down on the `Cloud CI 1` failure — it isn't this PR's.

- **Failing check:** `Cloud CI 1` (`.github/workflows/test-cloud-prod.yml`), run [34059719616 job 101557956680](https://github.com/VH-Lab/NDI-matlab/actions/runs/34059719616/job/101557956680).
- **Failing test:** `ndi.unittest.cloud.FileSeriesRoundTripTest/testManifestSurvivesADownloadFromTheCloud`, erroring inside `did.database/add_docs` (via `ndi.cloud.downloadDataset` → `ndi.dataset.dir` → `database.add`) with *"Add the document that authored the series, or re-author it with addFileSeries."*
- **Why it isn't this PR's:** the same job is red on `main` at the base commit `f0657312`, run [34059245139 job 101556682557](https://github.com/VH-Lab/NDI-matlab/actions/runs/34059245139/job/101556682557), with the identical `60 Passed, 1 Failed, 1 Incomplete` totals. This PR's only change is a 11-line dead-code deletion in `download_file_from_cloud`'s handler prologue in `didsqlite.m` — that code path is downstream of `ndi.cloud.downloadDataset`'s `add_docs` call site and is not reached before the failure.
- **Fix:** none exists yet — this looks like the fileSeries manifest / parent-document scoping issue on cloud round-trip and needs its own investigation. Not widening this PR to chase it.

`Cloud CI 2`, `Cloud CI 3`, No Cloud, and Symmetry are green here on the head commit; I'll keep the PR watched until it's mergeable.

---

stevewds (approved):

PR #965, opened 2026-09-06
Fix #964: retire uploadToNDICloud and scanForUpload; newDataset delegates to uploadDataset

### LINKED ISSUE #964: Not urgent: retire uploadToNDICloud / scanForUpload — unreachable in-tree, broken on file_path, and blind to file series
**Not urgent.** Nothing in the repo reaches this code, so nothing is currently failing because of it. Filing it so the decision is recorded rather than rediscovered.

Found while fixing the upload half of #956. The short version: there are two upload paths, only one is maintained, and the other appears to have been left behind when the file manifest contract changed.

## The two paths

| | maintained | this one |
|---|---|---|
| entry | `ndi.cloud.uploadDataset` | `ndi.cloud.upload.newDataset` |
| files | `uploadFilesForDatasetDocuments` → `ndi.database.internal.list_binary_files` | `uploadToNDICloud` → `ndi.cloud.upload.scanForUpload` |
| then | `zipForUpload` | `zipForUpload` |

Both producers feed the same `zipForUpload`, and both walk `files.file_list` with the same legacy `NAME#` expansion.

## Three problems, in order of how much they matter

**1. It looks broken.** `zipForUpload` reads each entry's `file_path`:

```matlab
current_file = files_to_process(i);
file_path = current_file.file_path;
```

`scanForUpload` never sets it — it populates `name`, `docid`, `bytes`, `is_uploaded`, `uid`, and nothing else. `base_dir = fullfile(D.path, '.ndi', 'files')` at `zipForUpload:41` is the only plausible fallback and is never used again in the file. So on any dataset with at least one file this should raise *"Reference to non-existent field 'file_path'"* as soon as the batching loop runs. A dataset with no files would pass, because the loop doesn't execute.

`list_binary_files` does set `file_path`, which is what suggests the contract changed and only the maintained producer was updated.

**2. It is blind to file series.** Same blind spot fixed in #961 for `list_binary_files`: a new-style series records members in a manifest rather than in `file_list` and names them `NAME_<i>`, which the `NAME#` branch never matches. A series would upload as its manifest alone.

**3. It has no test coverage**, which is how 1 and 2 both survived.

## Reachability — and a correction

I said on #956 that this was a "live, user-facing route" via the metadata app. **That was wrong**, and I'd rather correct it here than leave it standing:

- `ndi.database.metadata_app.fun.submit_dataset` calls `uploadToNDICloud`, but only *after* an unconditional `error('...is deprecated. Use ndi.cloud.uploadDataset instead')` on the line above it. The call is unreachable. Nothing calls `submit_dataset` either.
- `ndi.cloud.upload.newDataset` calls `uploadToNDICloud` and has **no callers anywhere in the repo**.

So the whole chain is dead in-tree. It is still *public API*: `newDataset` carries a worked example in its help, and `uploadToNDICloud` is documented in `docs/NDI-matlab/manual/developer/cloud.md`. That is the only reason this needs a decision rather than a delete.

## Suggested resolution: retire it

`newDataset`'s only distinct behaviour is steps 1a/1b — build the metadata structure and create the cloud dataset record — after which it just uploads. That part can stay and hand off to `ndi.cloud.uploadDataset`, leaving one upload path to maintain and test.

That would mean deleting `uploadToNDICloud` and `scanForUpload`, repointing `newDataset`, deleting the unreachable tail of `submit_dataset` (or the function, since nothing calls it), and updating the developer manual.

The alternative — repair `scanForUpload` to the `file_path` contract and teach it about series — means maintaining and testing two producers of the same structure forever, and the second one has already drifted once without anyone noticing.

**Deliberately not done as part of #961**, which stayed confined to the live path.

⚠️ Static reading only; I have not run `newDataset`. The three facts behind claim 1 (`zipForUpload` reads `file_path`, `scanForUpload` never sets it, `base_dir` is unused) are each verifiable by inspection, but the runtime behaviour is inferred, not observed.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

stevevanhooser:
Fixes #964. Two upload paths existed; only one was maintained. This removes the other rather than repairing it.

## Why removal, not repair

Both producers fed the same `zipForUpload`, so keeping both means maintaining and testing two producers of one structure. The unused one had already drifted without anyone noticing — **twice, independently**:

**1. `scanForUpload` doesn't satisfy `zipForUpload`'s contract.** `zipForUpload` reads each entry's `file_path`; `scanForUpload` writes `name`, `docid`, `bytes`, `is_uploaded`, `uid` and never `file_path`. `base_dir` at `zipForUpload:41` is the only plausible fallback and is dead code. Any dataset with at least one file would raise *"Reference to non-existent field 'file_path'"* once the batching loop ran.

**2. `newDataset` never even got that far.** It asks `createCloudMetadataStruct` for three outputs:

```matlab
[status,response,datasetId] = ndi.cloud.utility.createCloudMetadataStruct(metadata_structure);
```

That function declares **one** (`function result = createCloudMetadataStruct(metadata_struct)`) — its help still documents a three-output form the signature lost. So `newDataset` failed at its second statement, before reaching the `file_path` problem.

Either alone makes the path unusable. Both at once, unnoticed, is what "unmaintained" means.

It was **also** blind to file series, exactly as `list_binary_files` was before #961: a new-style series names members `NAME_<i>`, which the legacy `NAME#` branch never matches, so a series would upload as its manifest alone. That's a third repair this path would have needed to be worth keeping.

## What replaces it

`ndi.cloud.uploadDataset` already does everything `newDataset` did, correctly. When a dataset has no remote it builds metadata with the same `ndidataset2metadataeditorstruct` call, converts it with `createCloudMetadataStruct` (single-output form), creates the cloud record, adds the remote-dataset document, then uploads through the maintained manifest.

`newDataset` becomes a wrapper over it — keeping the documented signature working, and working for the first time in a while — with its help pointing at `uploadDataset` for new code. It raises on failure rather than returning a flag, since that's what its single-output signature can express.

## Also

- `submit_dataset`'s body below its unconditional deprecation `error(...)` is removed. It called `uploadToNDICloud` and could never run; leaving it implies a working path. The function still errors, since it's public and deprecated rather than absent, and nothing calls it.
- Developer cloud manual: `uploadToNDICloud` and `scanForUpload` sections dropped, `newDataset`'s notes the delegation.
- The stale `vltInNDI.md` row for the deleted file is removed.

## Testing

**No tests referenced any of this** — which is how it stayed broken through two contract changes. Nothing is lost by deleting it, and nothing new is testable without cloud credentials: `newDataset` is now three lines of delegation, and the code it delegates to is already exercised by Cloud CI.

Reachability, for the record: `submit_dataset` errors before its call and has no callers; `newDataset` has no callers in the repo. The path survives only as public API surface, which is the sole reason this is a retirement rather than a deletion.

⚠️ Static reading — I have not run `newDataset` against the cloud. The facts behind each breakage are verifiable by inspection; the runtime behaviour is inferred.

---


stevevanhooser:
**`codecov/patch` is red, and I am not adding a test to move it.** Recording why once.

Every other check is green: `Run tests (MATLAB latest)` (1,241 tests, 1,208 pass, **0 fail, ±0 test count** — nothing referenced the deleted files), `Test Results`, `Code Analyzer`, `Run symmetry tests`, `Cloud CI 1`, and `codecov/project`.

The patch is a deletion plus a three-line delegation:

```matlab
[success, datasetId, message] = ndi.cloud.uploadDataset(D);
if ~success
    error('NDI:Cloud:Upload:NewDatasetFailed', 'Failed to upload the new dataset: %s', message);
end
```

Those lines cannot be covered without cloud credentials — `uploadDataset` reaches the network on any input this repo's fast suite can build. I checked whether the failure branch could be exercised offline: `uploadDataset`'s first gate is `ndiDataset.isIngested()`, and `ndi.dataset.isIngested` documents that *"a dataset with no sessions is considered ingested"* and returns true, so an empty dataset goes straight past it. Reaching a `success == false` return without a network would mean constructing a dataset with a linked, non-ingested session from example data — a heavy fixture whose subject would be `uploadDataset`'s gate, not this delegation.

A test like that would raise the number and test nothing this PR changed. The code it delegates to is already exercised by Cloud CI.

Worth being plain about the coverage direction, since it cuts both ways: this PR **deletes** two functions that had no tests at all, which is why `codecov/project` improves while `codecov/patch` reports on the handful of new lines. The untestable-in-CI surface is smaller after this change than before it, not larger.

---

stevewds (approved):

PR #967, opened 2026-09-06
Read a series member's bytes back out of the cloud (#966)

### LINKED ISSUE #966: No test reads a file series MEMBER back out of the cloud: the round trip is only proven for the manifest
`ndi.unittest.cloud.FileSeriesRoundTripTest` is the only end-to-end test of the file series cloud path, and it never reads a member's bytes out of a downloaded dataset. What it actually covers today:

| test | reads from | covers |
|---|---|---|
| `testManifestSurvivesTheRoundTrip` | `LocalDataset` | upload accepted |
| `testDocumentReportsItsSeries` | `LocalDataset` | local document shape |
| `testCurrentFileListDoesNotExpandTheSeries` | `LocalDataset` | local file list |
| `testMembersSurviveTheRoundTrip` | `LocalDataset` | **members read locally, not from the cloud** |
| `testManifestSurvivesADownloadFromTheCloud` | downloaded copy | manifest bytes + uids + `seriesCount` |

`testMembersSurviveTheRoundTrip` is named for a round trip but asserts against `testCase.LocalDataset`, which is never re-downloaded — the members it verifies are the same bytes the test wrote to disk in setup. So a server that accepted member uploads and then dropped, truncated, or swapped their bytes would pass the whole class.

The consequence is that #961 (upload enumerates series members) and #959 (download reconstructs member locations) are both green without anything having proved that a member's bytes survive the journey. The batch presign path added in DID-matlab#183/#187 is likewise covered only by the offline `BatchSignedUrlLookupTest`; nothing exercises it against the real API.

### What is needed

Add a test to `FileSeriesRoundTripTest` — alongside `testManifestSurvivesADownloadFromTheCloud`, reusing the same download fixture — that opens a member on the **downloaded** dataset and compares its bytes to `testCase.MemberContent{i}`:

```matlab
fobj = downloaded.database_openbinarydoc(remoteDoc, sprintf('chunkdata.bin_%d', i));
got  = fread(fopen(fobj.fullpathfilename), Inf, 'uint8=>uint8');
```

Worth covering in the same test, since they are the failure modes a manifest-only check cannot see:

- every member present, not just member 1 (a per-member loop, so a partial upload shows which slot);
- **bytes compared, not sizes** — members in the fixture are written at distinct lengths, but equal-length wrong bytes is exactly the swap a uid mixup produces;
- a sparse series, if the fixture can carry one: an absent slot must come back absent rather than as a zero-byte file.

### Blocked on DID-matlab#188

This cannot be written today. `did.implementations.sqlitedb.seriesMemberPath` never retrieves a member whose bytes are not already on the machine: it resolves the manifest (retrieving it through the `customFileHandler` if needed) and then looks the member up in the local cache, returning `''` when it misses. A member carries no location of its own, so a freshly downloaded dataset that synced only the manifest has no way to reach one, and `database_openbinarydoc` on `chunkdata.bin_1` fails.

DID-matlab#188 adds that retrieval — the handler is called for a member with `mode` `'open'`, given the manifest's location and the member's uid. Once it lands and NDI's handler answers for members, this test becomes writable, and writing it is also what would first exercise the batch presign path end to end.

Related: #956 (the upload/download gaps this test would have caught), #961, #959.

stevevanhooser:
Fixes #966.

**A file series member's bytes now make the full round trip, and something checks that they do.** Cloud CI 1 at `504c5bc`: **62 passed, 0 failed, 0 incomplete** — the partition holding `FileSeriesRoundTripTest`, which had been red on every prior run of this branch.

The earlier do-not-merge warning is gone: it applied while this PR was deliberately red pending VH-Lab/DID-matlab#188, which has since merged along with DID#191.

## The problem

`FileSeriesRoundTripTest` had five passing tests and did not establish that a member's bytes survive the trip to the cloud and back.

`testMembersSurviveTheRoundTrip` was named for a round trip and asserted against `testCase.LocalDataset`, which is never re-downloaded — it compared the same bytes setup had written to disk a moment earlier, with no transfer in between. The only test that downloaded, `testManifestSurvivesADownloadFromTheCloud`, read the manifest and the counts. So a server that accepted the member uploads and then dropped, truncated, or swapped their bytes passed the entire class. That is how #961 (upload enumerates members) and #959 (download reconstructs locations) both went green with nothing having shown a member makes it there and back.

## What is here

**The test.** `testMembersSurviveADownloadFromTheCloud` downloads into a fresh folder and, for every member: opens it on the **downloaded** dataset and compares its bytes **in full**. Setup gives all four members the same 64-byte length and distinguishes them by content, so a size check would catch nothing — and a uid mixup, the failure this path is most prone to, produces a right-sized wrong file.

Each open is caught per slot rather than allowed to abort the loop, because *which* members came back separates bugs with different owners. The `exist_doc` probe is recorded as a diagnostic, not a gate: `check_exist_doc` calls `seriesMemberPath` without `mayRetrieve` by design, so a downloaded member reads as absent there even when perfectly fetchable. Asserting on it would have reported a working feature as a DID bug.

**Two source fixes**, both found by running that test rather than by reading:

1. `download_file_from_cloud` took the uid from the `ndic://` sourcePath. For a member that path names the **manifest**, so it fetched the manifest again and would have cached it under the member's uid — every later read of that member returning the manifest, quietly and permanently. DID's `HandlerReturnedManifest` guard caught it, which is why nothing was ever corrupted while the repos were out of step. `readContext` now returns `uid` and a member fetch uses it, only when `seriesName` marks it as a member: DID passes a uid on ordinary opens too, and the `ndic://` uid is the only identifier the cloud can mint a URL for.

2. A downloaded series manifest now keeps **both** locations — the local copy and the `ndic://` reference under the uid the cloud knows it by. DID#191 made DID offer a locally-synced manifest to the handler, which was necessary but not sufficient: the location it offers is a local path, and `download_file_from_cloud` raises `UnsupportedFileLocationType` for anything that is not `ndic://`. DID treats a failing handler as a plain miss, so members stayed unreadable and Cloud CI failed *identically* before and after #191 landed.

**Offline coverage** for that second fix, in `TestDownloadedSeriesManifestKeepsItsCloudLocation`: the local copy survives, the reference carries the original uid and is lazy (`ingest` 0, `delete_original` 0), a plain file gets no second location, and no cloud dataset id yields no reference rather than `ndic:///<uid>`.

**Dev-environment runners.** `testCloudApiProd` had no dev sibling; `testCloudApiDev` mirrors it, and `testFileSeriesCloud` runs one cloud class against a chosen environment (default: the file series round trip against dev). It refuses to start without credentials, because without them the cloud tests skip themselves and report Incomplete rather than Failed — a quiet run that reads like success.

**Housekeeping.** `testMembersSurviveTheRoundTrip` → `testMembersAreIngestedLocally`, which is what it does, with its stale skip gate dropped (member ingestion landed in DID#183). The class header, which still listed member ingestion and series-aware URL signing as outstanding, now says which tests read from the cloud and which do not — the naming is what hid this.

## What this still does not prove

A green run does **not** show the batch presign path was used. `batchSignedUrlLookup` returns `""` on any failure and `download_file_from_cloud` falls back to per-uid `getFileDetails`, which resolves members perfectly well. And the `fileSeries` scope is a scale property — at four members, scoped and unscoped are both correct and both fast, so no assertion at this size can tell them apart. Filed as #968.

## Status

Cloud CI 1 and 2 green at `504c5bc`; CI 3 still running. The head `5cca2f7` adds only the two task files under `tools/tasks/`, which no test path touches. Code Analyzer and symmetry green on the head.

### DIFF
--- a/src/ndi/+ndi/+cloud/+sync/+internal/updateFileInfoForLocalFiles.m
+++ b/src/ndi/+ndi/+cloud/+sync/+internal/updateFileInfoForLocalFiles.m
@@ -68,6 +68,37 @@ function document = updateFileInfoForLocalFiles(document, fileDirectory, cloudDa
             filename = originalFileInfo(i).name; % name for ingestion
             if isfile(file_location)
                 document = document.add_file(filename, file_location);
+
+                % A SERIES MANIFEST also keeps the cloud reference it came
+                % from, as a second location alongside the local copy.
+                %
+                % Its members are still on the cloud -- they are left there
+                % deliberately, so that opening a dataset does not drag down
+                % a 28,000-member series -- and a member has no location of
+                % its own. DID resolves one by handing the MANIFEST's
+                % location to the customFileHandler with the member's uid in
+                % the context, so that location is the only thing telling the
+                % handler where to look. With the local path alone,
+                % download_file_from_cloud is handed a path that does not
+                % start with 'ndic://' and raises
+                % NDI:Didsqlite:UnsupportedFileLocationType, which DID takes
+                % as a plain miss -- so every member of a downloaded series
+                % is unreadable, silently. See VH-Lab/NDI-matlab#966.
+                %
+                % The uid in the reference is the file's ORIGINAL uid, the
+                % one the cloud knows it by. add_file mints a fresh uid for
+                % this second location's own record, which is harmless: DID
+                % passes only the location STRING to the handler, and the
+                % member's uid comes from the context.
+                %
+                % Manifests only. Every other file is already here, and a
+                % second location on each would be rows to no purpose.
+                if strlength(cloudDatasetId) > 0 && document.isFileSeries(filename)
+                    % ndi.document.add_file recognizes 'ndic://' and supplies
+                    % ingest 0, delete_original 0, location_type 'ndicloud'.
+                    document = document.add_file(filename, ...
+                        sprintf('ndic://%s/%s', cloudDatasetId, file_uid));
+                end
             else
                 warning('Local file does not exist for document "%s"', ...
                     document.document_properties.base.id)
--- a/src/ndi/+ndi/+database/+implementations/+database/didsqlite.m
+++ b/src/ndi/+ndi/+database/+implementations/+database/didsqlite.m
@@ -166,13 +166,28 @@ function download_file_from_cloud(destPath, sourcePath, context)
     % CONTEXT is the per-call struct DID's dispatchCustomFileHandler passes
     % when a handler declares three or more inputs (DID-matlab#186): it
     % carries the caller's documentId and, for a series member, the
-    % series name. When present those are used to look the uid up in a
-    % per-document signed-URL cache -- one API round trip per DOCUMENT
-    % (or per fileSeries scope) instead of one per uid, closing DID
-    % #173's step 3 for the read path. Falls back to per-uid
+    % series name and the MEMBER'S OWN UID. When present those are used to
+    % look the uid up in a per-document signed-URL cache -- one API round
+    % trip per DOCUMENT (or per fileSeries scope) instead of one per uid,
+    % closing DID #173's step 3 for the read path. Falls back to per-uid
     % getFileDetails when no context is given or the batch call cannot
     % answer -- so a 2-arg handler-shaped caller, and any uid missing
     % from the batch response, still work.
+    %
+    % WHICH UID IS BEING ASKED FOR. On an ordinary file, SOURCEPATH names
+    % it and its uid is the one in the ndic:// reference. On a SERIES
+    % MEMBER it is not: a member has no location of its own, so DID passes
+    % the SERIES MANIFEST's location as SOURCEPATH and names the member in
+    % context.uid (DID-matlab#188), with context.seriesName marking that it
+    % is a member at all. Parsing the uid out of SOURCEPATH there
+    % fetches the manifest a second time and stores it under the member's
+    % uid, so every later read of that member returns the manifest --
+    % quietly, and forever. DID guards against exactly this, comparing what
+    % a handler returns against the manifest it already holds and refusing
+    % a match with DID:SQLITEDB:FileSeries:HandlerReturnedManifest, which is
+    % what this handler tripped on every member fetch before the context
+    % was read. So the member's uid comes from the context; every other
+    % call still takes it from sourcePath.
     %
     % Previously a nested function inside do_openbinarydoc. It is file-local so
     % that do_add can pass the same handler.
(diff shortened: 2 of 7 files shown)


stevevanhooser:
Cloud CI 1 is red, as the PR description said it would be, and the failure is the expected one — reporting once rather than pushing a fix, because the fix is not in this repository.

[Run 34063109469](https://github.com/VH-Lab/NDI-matlab/actions/runs/34063109469/job/101567067375) at `ed232d7`: **61 passed, 1 failed**. The single failure:

```
ndi.unittest.cloud.FileSeriesRoundTripTest/testMembersSurviveADownloadFromTheCloud
  Failed by assertion, at FileSeriesRoundTripTest.m:484

  a series member of the downloaded dataset could not be reached.
  "Members reachable in the downloaded dataset: 0 of 4"
  APIResponseBody: [false, false, false, false]
```

**Zero of four**, which is the signature the test was built to distinguish. Per the per-member check: none reachable is VH-Lab/DID-matlab#188 — nothing asks the `customFileHandler` for a member's bytes at open time — while *some* reachable would have meant the transfer lost members, which would be NDI's bug and mine to fix here. It is the former. The upload and download both did their jobs; the document, its manifest and its counts all came back (`testManifestSurvivesADownloadFromTheCloud` passes in the same run), and the members are on the cloud with nothing on this side able to ask for them.

No re-run: this is deterministic, not a flake, and re-running would burn the allowance to watch it fail identically.

Pushed two commits since that run, neither of them an attempt at the failure:

- `5bdc45c` — folded in the review note on #966 (comments only). It caught a real overclaim: I had marked class header item 3 DONE on the strength of the upload half, when no member uid has ever been through `batchSignedUrlLookup` on the way *back*. Now HALF DONE, and named as the second suspect if this test still fails once #188 lands.
- `123ce6d` — the run printed `slots zeros(1,0)`, `mat2str`'s rendering of an empty row, in the one line whose job is reporting the zero case and in a diagnostic written for readers who are often not MATLAB users. Now says `none`.

DID-matlab#188 is in progress. This PR should go green on its own once it merges and CI resolves the new DID; I am watching both and will refresh the branch when it lands.

---

stevevanhooser:
Still red at `9137663`, but for a **different reason than before**, and this one is worth stopping on. 61 passed, 1 failed. No Cloud green (1,279 tests, 0 failed), Symmetry green.

```
Members already in the local cache before any open: none
Members successfully opened: 0 of 4, slots none
First open failure: DID:SQLITEDB:open: The file "chunkdata.bin_1" ... its bytes
are not on this machine and could not be retrieved.
```

**No `HandlerReturnedManifest` warning anywhere in the run.** DID#188's guard fires when a handler is called and returns the manifest's bytes; its absence means the handler was never called at all. So the context-uid fix in `45fbaa7` is not what is failing — nothing got as far as using it.

### Why the handler is never asked

`fetchSeriesMemberBytes` looks up the series manifest's row for a location to offer, and only offers it when the location's type is **not** `'file'`:

```matlab
isEligible(idx) = ~strcmp(thisType, 'file');
if ~any(isEligible), return, end
```

Its rationale, from the source: *"a manifest that is local is also a database with no remote store to fetch a member from, so there is nothing to gain by trying."*

That assumption does not hold for `ndi.cloud.downloadDataset`. With `SyncFiles`, `updateFileInfoForLocalFiles` re-adds every document file by a plain local path:

```matlab
file_location = fullfile(fileDirectory, file_uid);
document = document.add_file(filename, file_location);   % -> location_type 'file'
```

so the downloaded database holds the manifest as `type = 'file'` at a local path — **while its members are still on the cloud and have never been fetched.** Local manifest, remote members is not a degenerate case here; it is the normal, intended shape of a lazily-synced dataset, and the exact configuration the whole file-series design exists to serve. DID reads that as "no remote store" and declines to ask.

### Two ways out

**Relax the rule in DID (recommended).** Its stated purpose has two halves, and both have expired:

- *"handing a local path over risks a handler copying the MANIFEST's bytes into the member's cache slot"* — that is precisely what `HandlerReturnedManifest` now catches, byte-for-byte, and refuses. The type rule is a second guard against a failure the first one already covers.
- *"a local manifest means no remote store"* — false for every dataset NDI downloads.

A handler that takes context (already required a few lines above) is claiming it can resolve a member from a uid. Letting it try, with the byte-comparison guard behind it, costs one refused fetch in the worst case.

**Or work around it in NDI.** DID inserts one `files` row per location and iterates every eligible row, so recording a second `ndic://` location on a downloaded series manifest alongside the local one would give `fetchSeriesMemberPath` something eligible to offer. That is defensible on its own terms — after a sync the manifest genuinely does have both a local copy and a cloud original — but it adds a location to work around a rule rather than fixing the rule.

I have not pushed either; the first is DID's call and the second shouldn't be done if the first is happening. Flagging rather than guessing.

Also on this head: `codecov/patch` reports 0% of diff hit, 6 lines in `didsqlite.m`. Those lines are inside `download_file_from_cloud`, a file-local function reachable only through DID with a live cloud dataset, so the only thing that can cover them is the Cloud test in this PR — which is the test that cannot run yet. It should resolve itself once this is unblocked.

---

stevewds (approved):

PR #969, opened 2026-09-07
Insist the batch presign path is used, and say so when it is not (#968)

### LINKED ISSUE #968: Nothing proves the batch presign path is ever used: the per-uid fallback makes success indistinguishable from N round trips
`ndi.cloud.download.internal.batchSignedUrlLookup` exists so that opening a 28,000-member series costs one API call per document (or per series scope) instead of 28,000 — DID-matlab#173 step 3, tracked here as #952. Nothing verifies it is used against a live API, and by construction nothing currently can.

## Why a green test says nothing

`download_file_from_cloud` tries the batch lookup first, then falls back:

```matlab
fileUrl = ndi.cloud.download.internal.batchSignedUrlLookup( ...
    string(cloudDatasetId), docId, seriesName, string(ndiFileUid));

if strlength(fileUrl) == 0
    % No batch URL for this uid -- either no document context, or the batch
    % call failed, or the returned map did not name this uid.
    [success, answer, ~] = ndi.cloud.api.files.getFileDetails(cloudDatasetId, ndiFileUid);
    ...
end
```

`batchSignedUrlLookup` returns `""` on *any* failure — network, auth, an unrecognized parameter, an unexpected response shape — and its own header says so: *"The batch is an optimization, not an authority."* That is the right runtime behaviour and exactly the wrong testing property. Since #961 uploads members as ordinary uid-addressed files, `getFileDetails` resolves any member perfectly well, so a run is green whether or not the endpoint ever answered.

There are at least two ways #967's round trip can pass, and it cannot tell them apart:

1. The endpoint ignores the `fileSeries` parameter and returns the whole document's URL set — which contains the member uids, since they are files of that document. Batch works, unscoped.
2. The endpoint rejects or mishandles it, the lookup returns `""`, and N `getFileDetails` calls quietly do the job.

## And the fixture cannot show the difference anyway

`FileSeriesRoundTripTest` uses **4 members**. The `fileSeries` scope exists so a 28,000-member series does not pull 28,000 URLs in one response — a scale property. At four members, scoped and unscoped are both correct and both fast. No assertion about bytes, timing, or correctness can distinguish them at this size.

So this is not "add an assertion to the existing test". It needs a test whose subject is *which calls were made*.

## What would actually verify it

**Offline, and worth having first.** `batchSignedUrlLookup` already takes an injectable `signer` and a `clearCache` flag, so a test can assert the shape of the calls without a server:

- one `signer` invocation serves N uids in the same scope (the cache does its job);
- a second scope triggers a second invocation, and the TTL expiry triggers a refetch;
- `seriesName` non-empty produces a `'fileSeries'`-scoped call, empty produces a document-scoped one.

`BatchSignedUrlLookupTest` covers some of this already — the gap is asserting the *scoping* and the call count, not just that a URL comes back.

**Against a live API.** The honest check is a call count: open every member of a series and assert `getFileDetails` was called zero times, or that the batch signer was invoked once. That needs a seam — an injectable signer threaded down to the handler, or a counter the test can read. `download_file_from_cloud` is a file-local function in `ndi.database.implementations.database.didsqlite` with no injection point today, so this is the part that needs design rather than just a test.

**Worth deciding explicitly:** whether a batch failure should ever be *loud*. Silent fallback is right for a user opening a file. For a 28,000-member series it converts a working feature into 28,000 API calls with no signal that anything regressed. A warning on first fallback per scope, or a counter surfaced somewhere, would make the difference visible without changing the behaviour.

## Context

Now runnable against dev: the API work this covers is deployed to `dev-api.ndi-cloud.com`, and #967 adds `testCloudApiDev` and `testFileSeriesCloud`, which run one cloud class against a chosen environment.

Related: #952 (the batch lookup), #966 (the round trip that surfaced this — same blind-spot shape, one layer up), #967.

stevevanhooser:
Fixes #968. Follows #967, which proved a series member's bytes make the round trip but could not say *how* they got there.

## The problem

`download_file_from_cloud` falls back to per-uid `getFileDetails` whenever `batchSignedUrlLookup` cannot answer. That is right at runtime — a reader opening one file should not fail because a batch endpoint hiccuped — and exactly wrong as a test property. `testMembersSurviveADownloadFromTheCloud` passed without anything showing whether the batch path answered or four per-uid calls quietly did the same work. At four members those are indistinguishable. At 28,000 they are one API call versus 28,000.

## Three changes

**1. Counters, so the fallback is visible.** `batchSignedUrlLookup` returns a second output: endpoint calls, uids answered from a batch, uids that fell back, and the last map's size and keys. Reading them with an empty documentId returns early, so a test can inspect them without disturbing the cache it is reporting on.

**2. The round trip now requires the batch path.** It clears the counters before opening members, then requires `uidMisses == 0` and `signerCalls == 1` — one presign call for the whole series, the rest from the in-process cache.

**3. A warning when it does fall back**, once per scope, `NDI:Cloud:BatchPresign:FallbackToPerUid`.

That third one is the change worth arguing for. The fallback produces no error and no log line, because the bytes do arrive. Reading a 10,000-member series then costs 10,000 presign calls, and what the user sees is not a failure but NDI being slow — so they conclude the tool is like that, and never report it. A single line naming the scope turns "this is slow" into "this fell back, and here is where." Once per scope rather than per uid, since the case worth warning about is exactly the one that would otherwise print 10,000 times.

## The scope question, which the counters cannot answer

`uidMisses == 0` catches a fallback. It does **not** catch a server that ignores the `fileSeries` scope and returns the whole **document's** URL set — that map contains the member uids and answers every one of them, so it looks identical to success. This is the live prod-vs-dev question: the `fileSeries` work is deployed to dev and possibly not to prod.

The only thing separating the two is *what else* the map names, and the existing fixture cannot pose the question — its document holds one manifest and four members, so scoped and unscoped differ by a single entry.

So this adds a document class, `demoNDISeriesMixed`, declaring three sidecar files alongside the series, and a test that fetches a **member** and requires the resulting batch map to name none of the sidecars' uids. If prod ignores the scope and dev honors it, this test fails on prod, passes on dev, and names the sidecars that leaked.

## Scope of the change

Nothing existing is modified. New class, new test method, plus the counters and the warning. The new test adds a second document to the dataset the fixture already uploaded rather than standing up its own, so it shares the cloud dataset and teardown and costs one extra upload instead of a second full lifecycle.

Counters are covered offline in `BatchSignedUrlLookupTest` against a scripted signer, including the miss case and the empty-documentId read — untested counters would relocate the blind spot rather than close it.

## Worth knowing when reviewing

The `demoNDISeriesMixed` class is the main risk: I cannot run MATLAB, so it is modeled on `demoNDISeries` and discovered by name from `ndi_common/` with no registry to update. A schema-loading mistake would surface as a failure in the new test.

The warning may also fire in other cloud tests that legitimately fall back. It is a warning, not a failure, so nothing breaks — and if it turns up as noise, that is information about how often the batch path is not answering.

---


stevevanhooser:
Cloud CI 1 is red at `d340268`, and that red **is the finding this PR was written to produce** — not a defect in it. Recording it so a reviewer does not misread the check.

`62 passed, 1 failed`. The single failure is `testMembersSurviveADownloadFromTheCloud`, at the new `uidMisses == 0` assertion. Every other assertion in that test passes: all four members open and their bytes match. The batch presign path simply did not answer, and the members were resolved one `getFileDetails` call at a time.

## Both environments behave the same way

A local run against **dev** produced:

```
Batch presign: 4 endpoint call(s), 0 uid(s) answered from the batch,
4 fell back to per-uid getFileDetails. Last batch map held 0 uid(s).
signerCalls: 4, uidHits: 0, uidMisses: 4, lastMapSize: 0, lastMapUids: []
```

Prod fails the same assertion here. So the working hypothesis going in — that prod ignores the `fileSeries` scope while dev honors it — does **not** survive: dev is not returning a narrower map, it is returning nothing usable. A zero-size map means `getSignedURLSetAll` produced either `ok = false` or a payload NDI does not recognize as `answer.files` being a `containers.Map`.

That leaves two candidates, and they have different owners:

- the endpoint refused or errored → server side;
- the endpoint answered and NDI failed to parse it → ours, in `signedURLFileMap` or `GetSignedURLSetAll`'s paging merge.

`d340268` could not tell them apart. Three commits since can.

## What the later commits add

- `a3ef0f6` — `stats.lastFailureReason` names which of the four causes it was, with the HTTP status when the signer reports one, and the warning and test narrative carry it. This is the line that decides where the bug lives.
- `aa0dc80` — two things a live run exposed. A failed scope was retried for **every** uid (four members, four signer calls), so an endpoint that cannot answer costs one failed batch call per uid *on top of* the fallback — 56,000 calls for a 28,000-member series, worse than having no batch path. Failed scopes are now remembered briefly. And `testTheBatchScopeIsTheSeriesAndNotTheWholeDocument` **passed vacuously** in this very run: an empty map leaks no sidecar uids, so it was satisfied by exactly the failure it should not survive. It now asserts the map is non-empty first.
- `4cbd893` — the fallback warning had no test, and shouted through three tests in its own offline suite. Both fixed.

The vacuous pass is worth dwelling on: it is the same shape as the bug this whole line of work started from in #966 — a test that cannot fail for the reason it exists. It took a live run against a broken endpoint to expose it.

---

stevevanhooser:
The blocker has changed, so an update on what this PR now waits for.

`5804cd8` switched the batch presign path to the `ndi-documents` routes from ndi-cloud-node#136. Cloud CI is still red, but the 404 it reports is a **different** 404:

| head | `lastFailureReason` |
|---|---|
| before | `the call reported failure (HTTP 404): Document not found` |
| `5804cd8` | `the call reported failure (HTTP 404)` |

The message is gone. That is a route 404 rather than a document 404 — prod does not have `/datasets/{id}/ndi-documents/{ndiId}/signed-url-set` yet, so the path itself 404s with no body. The NDI change is addressing the right endpoint; that endpoint is deployed to **dev** and not yet to prod.

**So this PR goes green when the routes reach prod, not when they reach dev.** `testCloudApiProd` is what Cloud CI runs. A dev deploy is enough to validate locally with:

```matlab
setenv('CLOUD_API_ENVIRONMENT','dev')
runtests('FileSeriesRoundTripTest.m')
```

Two other things this run showed, both working as intended:

- `testTheBatchScopeIsTheSeriesAndNotTheWholeDocument` now fails at the `assertGreaterThan(lastMapSize, 0)` guard added in `aa0dc80`, saying *"the batch returned no map at all, so nothing here can judge whether the 'fileSeries' scope was honored"*. Before that guard it **passed** in exactly this situation — an empty map leaks no sidecar uids — which is the vacuous pass the guard exists to stop. First live outing, and it did its job.
- The 404 arrives as one warning per scope naming the cause, rather than as unexplained slowness.

`5ec64d6c1` since fixes the one genuine regression here: `testSeriesNameIsPassedToTheSigner` asserted `args{1} == 'fileSeries'` positionally, which the new `idNamespace` pair displaced. Name-value order is not a contract; the pairs are found by name now.

---

stevevanhooser:
**Validated against dev.** @stevevanhooser reports the full cloud battery passing against the dev API with the `ndi-documents` routes deployed, including `FileSeriesRoundTripTest` in its entirety, and the offline suites passing locally. Prod is deploying now.

That closes the question this PR was opened to answer. The batch presign path had never worked — not on prod, not on dev, not since #952 — and nothing said so, because the per-uid fallback quietly produced correct bytes. A user would have seen a working system that was simply slow, and concluded that was how it worked.

## What the dev pass actually establishes

Three things that the pre-#968 test suite could not have told apart:

1. **The endpoint answers.** `signerCalls == 1`, `uidMisses == 0` — one batch call for the whole series, no member resolved by `getFileDetails`.
2. **The `fileSeries` scope is honored.** `testTheBatchScopeIsTheSeriesAndNotTheWholeDocument` passes with `lastMapSize > 0` *and* no sidecar uid in the map. The fixture (`demoNDISeriesMixed`) carries three non-series files precisely so a server that ignored the scope would leak them. It doesn't.
3. **The bytes still make the round trip.** `testMembersSurviveADownloadFromTheCloud` reads from the downloaded copy, not the local dataset it was written from.

Point 2 is worth separating from point 1: an endpoint that returns the whole document's URL set answers every uid, so it satisfies `uidMisses == 0` while doing the wrong thing. The map-size guard and the leak check are what distinguish them, and both had to be there — the leak check alone passed **vacuously** against a dead endpoint, since an empty map leaks nothing.

## Since the last comment

- `52f56c0f` — eight offline tests for the namespace routing, plus a `buildURL` seam on `CreateSignedURLSetJob`, whose URL was previously built inside `execute` and so unreachable without a live server. They check that `"cloud"` still selects the by-`_id` route (nothing moved for existing callers), that `"ndi"` selects `ndi-documents`, that `limit`/`cursor`/`fileSeries` survive the route change, that an empty NDI document id is refused rather than building `.../ndi-documents//signed-url-set`, and that a typo'd namespace throws at the call instead of defaulting into a 404.

  These exist because "it worked once on dev" is thin protection for the branch that took an evening to find. They run in the no-cloud suite, so a regression surfaces in minutes rather than at the next live cloud run.
- `52a2e280` — Code Analyzer, correctly: the `buildURL` extraction left `answer = []` dead, since the 202 check and its `else` both assign it.

## What remains

Only the prod deploy. Cloud CI runs `testCloudApiProd`, and the last red — `61 passed, 2 failed` at `5ec64d6c` — was the route 404 with no body, i.e. the path not existing rather than the document not being found. No further NDI change is needed for this to go green.

One incidental confirmation from that run: the negative caching in `aa0dc80` is working. Four members cost **one** signer call against the dead endpoint, where the earlier run made four. A missing endpoint no longer costs a failed batch call per uid *on top of* the fallback.

---

stevewds (approved):

PR #977, opened 2026-09-08
GEF Manager, an HDF5 table of contents, and one-based tile files

stevevanhooser:
Follow-on to #933. Five things: the app is renamed and can open a pyramid in napari, three of its table columns were reading from the wrong document, a new utility says what kind of HDF5 a file is, a source file can be *described* without being copied, and — the largest of them — tile files are renamed to match the convention DID enforces, which fixes an upload that was silently losing most of a pyramid.

## Tile files count from one, and the uploader stops guessing where a series ends

Two defects. The second is why the first mattered.

**The naming.** `did.document/addFileSeries` takes "ONE-BASED member numbers" and rejects anything else — *"Member indices must be positive integers (one-based)"*. `makePyramid` named its first tile `tile.bin_0`, putting every pyramid one step outside the convention its own storage layer assumes.

The tile **index** stays zero-based: it is a grid position, and `index_order` defines it as `row*tile_columns + column`. Only the file **suffix** moves, so the file holding tile N is `tile.bin_<N + tile_index_origin>`.

The origin is *recorded* on the tiles document, not inferred. Pyramids already built name their first tile `tile.bin_0`, carry no value, read as 0, and keep opening. Inferring it from the stored names cannot work: a sparse pyramid whose first tile is empty has no `_0` under either convention, and a wrong guess shifts every tile one cell along its row — and the last tile of a row into the next one. An image that is quietly wrong is worse than one that is missing.

**The walk.** `ndi.database.internal.list_binary_files` walked a legacy `NAME#` series as `NAME1, NAME2, …` and treated the first absent name as the end of the series, because that gap was the only termination signal a probing loop has.

A sparse series defeats that, and sparse is the normal case here: neither pyramid writer emits a tile with no data, so a level's stored tiles run 1, 4, 9, … with holes between. The walk found the first hole and called it the end.

One real pyramid reached the cloud as **24 of its 450 tile files**, with the upload reporting success. Documents and binaries travel separately, so the dataset downloaded and opened looking complete, then failed at the first tile read — an hour downstream of the cause, with a 404 that reads as a permissions problem.

`files.file_list` declares a *pattern*; `files.file_info` records the concrete names that were actually added, which is what `current_file_list` returns. Asking the document removes the guessing and the terminator together — there is nothing to count up to. The new-style `file_series` pass below it is unchanged and still needed: `addFileSeries` calls `add_file` once, for the series manifest, so `file_info` holds the manifest's name and not the members'.

## The rename

`ndi.gui.app.GeneIngest` becomes `ndi.gui.app.GEFManager`, menu label "GEF Manager". "Gene Ingest" named one of the four things the window does, and the list it opens on is of pyramids that already exist, so the name described a button rather than the window. It lists, views, adds and deletes.

The class is renamed rather than only its label: apps are discovered by scanning for `sessionApp` subclasses, so a deprecated forwarding shim would have appeared in the menu a second time under the same label — there is no compatible way to keep both.

## View

`View...` hands the pyramid to napari, which is Python, so nothing is drawn in MATLAB. `viewCommand` builds a command line; a dialog collects the options and runs it.

The launcher defaults to `/usr/local/bin/napariViewGEF` and is remembered in the new preference `GUI.GEFManager.ViewerLauncher`. That path is a shell wrapper rather than the console script directly, because MATLAB exports library paths of its own and a Python process started from MATLAB loads MATLAB's copies of libraries it must not use; the wrapper scrubs the environment before exec'ing the real entry point.

`viewCommand` is pure — no filesystem, no launch — so the exact command is on screen before it runs and stays there afterwards, and it is tested headlessly with no display and no Python. Launching another program is the step most likely to fail for reasons NDI cannot see, and a command the user can paste into a terminal is the difference between a shrug and something they can act on.

Two things the builder settles rather than leaving to chance: the pyramid is always named (a lone pyramid opens unprompted, but a session holding two makes the viewer list them and open nothing, which would look like View doing nothing at all), and `--outlines` implies `--cells` (outlines without cells is a refusal, not a picture with fewer layers). Paths and layer names are shell-quoted.

## Three columns were reading from the wrong place

| column | was | is |
| --- | --- | --- |
| Genes | `spatialGeneExpressionPyramid.n_genes`, which does not exist → `NaN` | the `geneList` document the pyramid depends on |
| Assay | `spatialGeneExpressionPyramid.assay`, which does not exist → blank | the `geneExpression` superclass property list |
| Subject | the subject's document id | its `local_identifier` |

`NaN` in a Genes column reads as "this pyramid has no genes" rather than "this code looked in the wrong place", which is why it went unnoticed. The id stays on the row as `subjectID` and is the fallback when a subject carries no `local_identifier` — an empty cell would claim the pyramid has no subject at all. Both lookups are one query for the session, not one per row.

## `ndi.util.hdf5Contents`

A Stereo-seq run leaves several files with the same stem and different middles — `Y00709C1E2.tissue.gef` beside `Y00709C1E2.adjusted.cellbin.gef` — and the extension does not say which is which. All of them are HDF5; `.gef` is a layout *inside* HDF5, not a format beside it, so "is this really an h5?" is always yes and never the question. The question is which groups it holds, because that is what decides whether `fromGEF` can read it.

`hdf5Contents` prints the tree and a verdict naming the layout: square-bin GEF, CellBin GEF, AnnData, or none of those. `maxDepth` and `maxChildren` keep an `.h5ad`'s thousands of `/obs` columns from being the whole output, and both say when they stopped rather than truncating silently.

## `fileReference`

A file the database *describes* but does not hold: filename, original path, format ontology, dates, size, checksum. `makeSourceFile` with `attachFile false` now produces one of these instead of a `generic_file` with no file attached — which was invalid, since `generic_file` requires `generic_file.ext`, and only became visible when DID-matlab#182 closed a fail-open.

## Also

`ndi.cloud.profile.reset()` now restores the detected backend. `profileTests` selected the in-memory backend and left the singleton on it for the rest of the session, so `setPassword` wrote to RAM with no error and `getPassword` handed it back — the tests passed and everything after them was talking to a store that persisted nothing.

## Tests

~50 new tests, all headless — no display, no Python, no network. The tile round-trip builds a pyramid whose 2×2 grid stores tiles 1 and 4 with nothing between them, then walks it the way the uploader used to and asserts the walk falls short of the stored set.

## Follow-on

The matching NDI-python changes are in Waltham-Data-Science/NDI-python#291: the same tile-naming change on the reader side, plus the napari panels and five cloud-credential fixes that stopped a downloaded dataset from opening at all. That PR also vendors `fileReference` into NDI-python's `ndi_common`, without which a Python reader could not validate a session ingested here.

https://claude.ai/code/session_01YHxsZ56GZsZ3hJ98TJ81t9


stevevanhooser:
## Confirmed on real data: 450 of 450

The PR body argues the bug from diagnosis. It has now been reproduced and fixed end to end on the opossum section, which is the dataset the 24-of-450 figure came from.

The section was rebuilt from `Opossum.tissue.gef` (9.41 GB, ~26,000 genes) with this branch checked out, via `buildOpossumDocuments` in `bscholl-genomics-python/ndiMatlabIntegration` — the same `fromGEF` / `fromCellBin` path as the app, without a display. Then, on a dataset holding that session:

```matlab
>> m = ndi.database.internal.list_binary_files(DS, docs, true);
Working on document 17 of 17
>> m
m =
  1×450 struct array with fields:
    name
    uid
    docid
    bytes
    file_path
```

**450.** Under the old walk the same section produced 24 — the first hole in a sparse level read as the end of the series, and the upload reported success. 450 is also exactly the number of `ndic://` uids the previously downloaded documents referenced, so the manifest is not merely larger, it is complete: every file the documents name is now in it.

Both halves of the change are exercised by that number. The tiles are one-based (`tile.bin_1` first, `tile_index_origin: 1` recorded on each tiles document), and `list_binary_files` enumerates `current_file_list()` rather than counting up to the first gap. Either fix alone would still have lost files: one-based naming with the old walk starts at a name it never probes, and the new enumeration over zero-based names would collide with the convention `addFileSeries` enforces.

The upload is running with all 450.

Worth noting what this does *not* retroactively fix, since it is the reason the bug was expensive to find: the previously uploaded dataset still holds 24 files. Documents and binaries travel separately, and `updateFileInfoForRemoteFiles` rewrites every location to `ndic://` without asking the server whether the uid exists — so that dataset downloads and opens looking complete and fails at the first tile read, with a 404 that reads as a permissions problem. Fixing the uploader does not move the missing bytes; the section had to be rebuilt and re-uploaded.

The unit test in this PR covers the same shape at small scale — a 2×2 grid storing tiles 1 and 4 with nothing between them, walked the way the uploader used to walk it, asserting the walk falls short of the stored set.

---

stevewds (approved):

Issue #986, opened 2026-09-14
Retire the manifest cloud-reference workaround and add an offline SyncFiles=false round-trip test once DID-matlab#201 lands


stevevanhooser:
## Context

VH-Lab/DID-matlab#201 (PR VH-Lab/DID-matlab#202, waiting on review) adds a lazy-fetch of a series manifest through the `customFileHandler`. Once it lands:

- A document downloaded with `SyncFiles=false` — manifest bytes never installed locally, only an `ndic://` location under `location_type='ndicloud'` — resolves members correctly. `sqlitedb.fetchSeriesManifestBytes` offers the manifest's `file_info` location to the handler, caches at `filecachepath/<manifestUid>`, and `seriesMemberPath` proceeds as normal.
- DID-matlab does NOT compose `ndic://` URLs or learn the scheme; NDI-matlab's handler does. So the "does DID call the handler correctly for a cloud-only manifest?" question is DID's to answer (VH-Lab/DID-matlab#202 pins that), and the "does the actual `ndic://` handler produce the right bytes when asked for a manifest by uid?" question is this repo's.

## What NDI-matlab already tests

- `ndi.unittest.cloud.FileSeriesRoundTripTest.testMembersSurviveADownloadFromTheCloud` — the acceptance test. Uploads a series, downloads with `SyncFiles=true`, opens each member. Requires NDI Cloud credentials (skips otherwise).
- `ndi.unittest.TestDownloadedSeriesManifestKeepsItsCloudLocation` — offline. Verifies that `ndi.cloud.sync.internal.updateFileInfoForLocalFiles` adds an `ndic://` location alongside the local manifest so members can be fetched. This is the workaround the DID-matlab#201 fix makes optional in the `SyncFiles=false` shape (the manifest is only remote) and reduces to "belt and suspenders" in the `SyncFiles=true` shape (a local manifest is offered to the handler too, since VH-Lab/DID-matlab#191).

Every existing round-trip test uses `SyncFiles=true`. `SyncFiles=false` appears only in `tests/+ndi/+unittest/+cloud/+sync/datasetDemo.m` (a demo, not a test).

## What is missing

[...]

PR #979, opened 2026-09-09
Add lightsheet OME-Zarr NDI document type + LightsheetZarrManager (scaffold)

stevevanhooser:
## Summary

Adds an NDI document representation for a lightsheet OME-Zarr (NGFF v0.4) multiscale pyramid, a builder that ingests one from an OME-Zarr store, and a GUI manager that hands a selected pyramid to napari through the same shell-wrapper pattern GEFManager uses.

Design mirrors `+ndi/+fun/+doc/+gene/` and `ndi.gui.app.GEFManager`, so a reader of the GEF ingest will recognise every piece here. Companion changes in NDI-python, NDR-matlab, NDR-python live on the same branch name in each of those repos.

### New documents
- `lightsheetZarrPyramid` — one per REDUCTION variant (`mean`, `max`). Carries axes, level-0 shape/voxel-size/translation, channels, dtype. Depends on subject + optional element + `fileReference` for the source store.
- `lightsheetZarrLevel` — one per (pyramid, level). Owns chunk bytes as a `chunk.bin_#` file series and the per-level metadata (shape, chunks, voxel size, translation, codec).

NDI documents are immutable, so a new reduction is a new pyramid + its own level ladder; a new level is a new level document.

### New MATLAB API
- `ndi.fun.doc.lightsheet.fromOMEZarr(session, zarrPath, ...)` — probe a store, write one pyramid per multiscales entry, one level per entry level. Metadata-only.
- `ndi.fun.doc.lightsheet.makePyramid`, `makeSourceFile`, `levelTable`, `chooseLevel`, `viewCommand` — pure/working helpers.
- `ndi.fun.doc.lightsheet.writeChunkFile`, `readViewport` — scaffolds; `raise notImplemented` today. See package README for the design and what's needed to finish them.
- `ndi.gui.app.LightsheetZarrManager` — session app that lists pyramids, adds from an OME-Zarr path, opens a pyramid in napari via `/usr/local/bin/napariViewLightsheet`, and cascade-deletes a pyramid with its levels.

### Companion repos on this branch name
- `VH-Lab/NDI-python` — `ndi.lightsheet.napari_view` console script + `PyramidReader`; `scripts/napariViewLightsheet` shell wrapper.
- `VH-Lab/NDR-matlab` — `ndr.format.omezarr.probe`, `ndr.format.omezarr.reduce`.
- `VH-Lab/NDR-python` — Python mirrors of the above two.

## Scope note

[...]

PR #988, opened 2026-09-14
Cloud companion: SyncFiles=false round-trip in FileSeriesRoundTripTest (#986)

stevevanhooser:
## Summary

Adds `testMembersSurviveADownloadFromTheCloudWithSyncFilesFalse` to `ndi.unittest.cloud.FileSeriesRoundTripTest`, closing the cloud-side coverage gap called out in VH-Lab/NDI-matlab#986. Companion to the existing `testMembersSurviveADownloadFromTheCloud` (SyncFiles=true) and to the offline-only #987 (mock handler).

## Why this test

Every existing cloud test downloads with `SyncFiles=true` and asks the handler only for **members**. VH-Lab/DID-matlab#202 (merged) opened a NEW code path: DID's `seriesMemberPath` now fetches the **manifest itself** through the customFileHandler on a local miss. That call shape — `ctx.uid = manifest's own uid`, `ctx.seriesName = ""`, `ctx.mode = "open"` — has offline coverage in #987 with a mock handler, but has never been exercised against NDI's real `download_file_from_cloud` in CI. This closes that gap.

## What the test does

Reuses the same upload/dataset lifecycle as `testMembersSurviveADownloadFromTheCloud`. Then:

1. Downloads the dataset with `SyncFiles=false` (no bytes installed, manifest arrives as ndicloud/ingest=0).
2. Verifies the precondition: neither the manifest nor any member is on this machine.
3. Clears the batch presign cache.
4. Opens every member end-to-end through `database_openbinarydoc` (which routes to `download_file_from_cloud`).
5. Verifies:
   - All members opened.
   - The manifest lands at `filecachepath/<manifestUid>` after the first open — the cache placement that makes every subsequent member open a `cached_path_for_uid` hit.
   - `batchSignedUrlLookup.uidMisses == 0` — every uid (the manifest's included) answered from the batch, no per-uid `getFileDetails` fallback.
   - Byte-for-byte member payloads against the uploaded content.

`signerCalls` is deliberately **not** tightly asserted: SyncFiles=false may reasonably cost one call (whole-document scope answers both manifest and members) or two (manifest, then series scope), and pinning one over the other would guess at a server behavior this suite does not otherwise cover.

## Why not the lightsheet fixture

`ndi.test.lightsheet.makeBlobFixture` (on the unmerged `claude/lightsheet-zarr-ndi-viewer-djp5vk` branch) writes a real OME-Zarr store and would exercise the whole ingest → upload → download → open pipeline. **Not used here** because:

- The branch isn't merged; basing this PR on it would either block on that or duplicate the fixture. The fixture's shape may still change before it lands.
- The question this test asks — *does `download_file_from_cloud` produce the right bytes when DID asks it for a manifest by uid?* — is narrower than a lightsheet ingest. `demoNDISeries` gives a direct byte-vs-byte comparison against known content, no extra moving parts.

A lightsheet-shaped end-to-end test (real OME-Zarr → `fromOMEZarr` → upload → `SyncFiles=false` download → open a chunk through `napariViewLightsheet` — the failure path VH-Lab/NDI-matlab#979 originally hit) is a natural follow-up once that branch lands.

## Related

- **VH-Lab/NDI-matlab#986** — the issue this and #987 together address.
- **VH-Lab/NDI-matlab#987** — the offline companion (mock handler, no cloud creds needed).
- **VH-Lab/DID-matlab#202** (merged) — the DID-side fix the test exercises.
- **VH-Lab/DID-python#88** (merged) — the Python companion port.
- **VH-Lab/NDI-matlab#979** / #966 — the original user-visible failure.

## Test plan

- [ ] `ndi.unittest.cloud.FileSeriesRoundTripTest.testMembersSurviveADownloadFromTheCloudWithSyncFilesFalse` — new. Requires NDI Cloud credentials (skips if `NDI_CLOUD_USERNAME` / `NDI_CLOUD_PASSWORD` are unset, same as its siblings).
- [ ] `testMembersSurviveADownloadFromTheCloud` — the SyncFiles=true sibling should keep passing (unchanged).
- [ ] `testTheBatchScopeIsTheSeriesAndNotTheWholeDocument` — unchanged.

Cannot run MATLAB locally in this session — please review and validate before merging.

https://claude.ai/code/session_015dKpDXCo8VTS4AHiibNnCp

---
_Generated by [Claude Code](https://claude.ai/code/session_015dKpDXCo8VTS4AHiibNnCp)_


stevevanhooser:
## CI status and finding

The first Cloud CI 1 failure was `installMatBox` timing out on `websave` — an infra flake, unrelated to this PR. Re-ran once, and this time it got past MatBox and reached actual test code. That surfaced a **real, pre-existing gap** in the `SyncFiles=false` download path that this test caught.

### What the new test found

`testMembersSurviveADownloadFromTheCloudWithSyncFilesFalse` doesn't even reach the manifest-fetch behavior it was designed to exercise. `ndi.cloud.downloadDataset(id, folder, 'SyncFiles', false)` errors during `ndi.dataset.dir`'s document ingest:

```
Error using did.implementations.sqlitedb/do_add_doc (line 450)
Refusing to add document 4126971f3c92265b_c097014ed0eb1479: its file series
"chunkdata.bin" declares 4 present members but records no location for any
of them.  A document read back from a database has had these stripped, so
it can be added to another branch of the database it came from but not to
a database that has never held it. Add the document that authored the
series, or re-author it with addFileSeries.
```

Stack: `ndi.cloud.downloadDataset` → `ndi.dataset.dir` → `ndi.database.add` → `didsqlite.do_add` → DID's `add_docs` → the DID-matlab#185 guard (`DID:SQLITEDB:FileSeries:MembersNotLocatable`).

### Root cause

The two download-path helpers differ in how they treat `series_info.ingest_locations`:

- **SyncFiles=true** calls `ndi.cloud.sync.internal.updateFileInfoForLocalFiles`, which — since NDI-matlab#958 — invokes `ndi.cloud.sync.internal.reconstructSeriesIngestLocations` to rebuild `ingest_locations` from the just-downloaded manifest file. That satisfies the DID#185 guard.
- **SyncFiles=false** calls `ndi.cloud.sync.internal.updateFileInfoForRemoteFiles`, which reshapes `file_info` to `ndicloud`/`ingest=0` but leaves `series_info.ingest_locations` empty. A series document with `n_present > 0` and empty `ingest_locations` is exactly what the DID#185 guard refuses.

So today a `SyncFiles=false` download of any dataset that contains a series document raises the guard's error before anything opens a member — regardless of what DID-matlab#201 landed on the read side.

### Proposed patch

The fix belongs in NDI-matlab, not in DID or this PR. Two options that both live under `ndi.cloud.sync.internal.*`:

1. **Have `updateFileInfoForRemoteFiles` reconstruct `ingest_locations` too.** For each series, fetch the manifest bytes on the fly (small — one uid_width record per slot, ~1 MiB for 28 000 members) via NDI Cloud, hand them to `reconstructSeriesIngestLocations`, then discard the local manifest copy afterwards to keep the "no bytes on disk" contract. Members stay remote; only the tiny manifest is pulled at download time to know which uids the series has.

2. **Have `downloadNdiDocuments` fetch manifests unconditionally.** `SyncFiles=false` becomes "no *member* bytes locally"; the manifest is always downloaded because it is what tells anything what members exist. Same net effect on total download size (manifests are already small), and DID-matlab#201's lazy-fetch of the manifest becomes redundant for the "downloaded via NDI Cloud" path but still matters for any other cloud-only ingest path.

Option 2 is smaller and matches the existing SyncFiles=true code path more closely; option 1 keeps the "manifest also stays remote" model DID-matlab#201 unlocked. Either fix is a small, self-contained change in `ndi.cloud.sync.internal.*` and doesn't require touching DID.

### What to do with this PR

Two ways to land the cloud test:

- **Fix the download path in this PR** (option 1 or 2 above), then the test passes end-to-end. Slightly widens PR scope beyond what NDI-matlab#986 asked for.
- **Land the test guarded**, so it currently reports the gap as `Incomplete` rather than a fail (`testCase.assumeError(@() ndi.cloud.downloadDataset(..., 'SyncFiles', false), 'DID:SQLITEDB:FileSeries:MembersNotLocatable')` around the download step, until the fix lands). That keeps this PR scoped to "add the cloud test skeleton" and files a separate NDI-matlab issue for the download-path fix.

I'll wait for the maintainer to pick between these.

For the record, the offline sibling PR (#987) is unaffected — its test uses `did.implementations.sqlitedb` directly and does not go through `ndi.cloud.downloadDataset`, so the DID#185 guard fires from a different site there and can't. `#987` is green after the closure fix.

---

stevevanhooser:
Pushed the fix (commit 1a6ec2f). Went with **option 1** per your call ("dynamically fetch the manifest — if we do it twice it will be in cache").

## What changed

`updateFileInfoForRemoteFiles` now rebuilds `files.series_info.ingest_locations` in the same shape `SyncFiles=true` produces (via the existing `reconstructSeriesIngestLocations`). For each series with `n_present > 0` and empty `ingest_locations`:

1. Look up the manifest's uid from `file_info` by series name.
2. Fetch the manifest bytes into a **scratch directory** via `batchSignedUrlLookup` (per-document scope) → `getFileDetails` fallback → `getFile`. Same mint‑URL‑then‑`getFile` didsqlite's own `download_file_from_cloud` runs for a non‑member ndic:// file.
3. Hand the scratch dir to `reconstructSeriesIngestLocations`; it reads each present slot's uid and builds the struct array.
4. Delete the scratch dir on exit (`onCleanup`).

Manifests do **not** land in `filecachepath` at download time. The first member open trips DID#201's `fetchSeriesManifestBytes`, which is what installs the manifest at `filecachepath/<uid>`. That preserves the test's precondition assertion (`assertFalse(isfile(cachedManifestPath))` before opening) and keeps the "if the cache is evicted, re-fetch works" property you flagged — the download path doesn't depend on any persistent local manifest file.

## Bounds

- No other caller of `updateFileInfoForRemoteFiles` in the repo; only `FileSeriesRoundTripTest.m` references it, and only in a comment.
- Docs with no series (or `n_present == 0` on every series) skip the whole block via the `needsReconstruction` gate — no extra API traffic.
- A manifest that cannot be fetched leaves the entry un-reconstructed; DID#185 then fires on `add_docs` with the document's own identity, which is the right voice for a partial download.
- No refactor of didsqlite's `download_file_from_cloud` — its 3‑arg handler contract and manifest‑vs‑member uid switch don't match what a download‑time caller needs, and lifting the shared piece into its own helper would widen this PR beyond the SyncFiles=false gap it closes. Kept the mint‑URL logic file‑local.

CI should now pass `testMembersSurviveADownloadFromTheCloudWithSyncFilesFalse` end-to-end.

---

stevewds (approved):

PR #989, opened 2026-09-14
Retire the manifest cloud-reference workaround (#986)

stevevanhooser:
## Summary

Closes the last item on VH-Lab/NDI-matlab#986. Removes the "belt and suspenders" workaround at `updateFileInfoForLocalFiles.m:96-101`, which used to write a second `ndicloud` location alongside the local `file` location on every downloaded series manifest.

## Why the workaround existed

Before DID-matlab#191, DID's read path could not reach a manifest that lived on the local disk directly — its customFileHandler only got called for non-`file` locations. So when `updateFileInfoForLocalFiles` re-pointed a downloaded manifest at its local copy, DID would try to open a member of the series, hand the local path to `download_file_from_cloud`, and hit `NDI:Didsqlite:UnsupportedFileLocationType`. DID treats a handler that fails as a plain miss (deliberately, since it resolves speculatively), so every member of a downloaded series read as absent — silently. See VH-Lab/NDI-matlab#966.

The workaround: keep the cloud reference alongside the local one, so the handler always had an `ndic://` path to work with, whichever DID reached for first.

## Why it can be retired now

Two DID-side fixes closed both sides of the gap:

- **VH-Lab/DID-matlab#191** — a local-file manifest is now offered to the handler as well, so DID reaches the local copy without the workaround having to route through `ndic://`.
- **VH-Lab/DID-matlab#201** (PR #202) — a cloud-only manifest resolves through the handler on a local miss. That path is exercised end-to-end by `testMembersSurviveADownloadFromTheCloudWithSyncFilesFalse` (VH-Lab/NDI-matlab#988, just merged).

The full cloud round trip now has coverage on both sides:

- `SyncFiles=true` → `testMembersSurviveADownloadFromTheCloud` (pre-existing).
- `SyncFiles=false` → `testMembersSurviveADownloadFromTheCloudWithSyncFilesFalse` (#988).

Either would fail if this retirement regressed member reads.

## What's in the diff

1. `src/ndi/+ndi/+cloud/+sync/+internal/updateFileInfoForLocalFiles.m` — remove the 30-line block that wrote the second location (both the code and the long comment explaining why the workaround existed). The `reconstructSeriesIngestLocations` call for DID#185's guard stays; that's unrelated.
2. `tests/+ndi/+unittest/TestDownloadedSeriesManifestKeepsItsCloudLocation.m` — **deleted**. Its whole subject was pinning the second location; with the workaround gone, its assertions describe behavior we deliberately no longer have.
3. `tests/+ndi/+unittest/+database/TestUpdateFileInfoPreservesSeries.m` — add `testSeriesManifestHasExactlyOneLocation`. Guards the new invariant: after `updateFileInfoForLocalFiles`, the manifest's `file_info` has exactly one location, and it's the local file. Reintroducing an `ndic://` second location trips this test immediately.

## Related

- VH-Lab/NDI-matlab#986 — the umbrella issue, closes with this PR.
- VH-Lab/NDI-matlab#987, #988 (merged) — the tests half of #986.
- VH-Lab/DID-matlab#191 — the earlier DID fix that made the workaround "belt and suspenders".
- VH-Lab/DID-matlab#201 / PR #202 — the DID fix that closed the last case (SyncFiles=false).
- VH-Lab/NDI-matlab#966 — the original silent-member-miss failure the workaround plastered over.

## Test plan

- [ ] `ndi.unittest.database.TestUpdateFileInfoPreservesSeries` — the new `testSeriesManifestHasExactlyOneLocation` should pass; the other three (from NDI-matlab#945) should keep passing.
- [ ] `ndi.unittest.cloud.FileSeriesRoundTripTest.testMembersSurviveADownloadFromTheCloud` (SyncFiles=true) — must keep passing after the workaround is gone. This is the real regression guard.
- [ ] `ndi.unittest.cloud.FileSeriesRoundTripTest.testMembersSurviveADownloadFromTheCloudWithSyncFilesFalse` — must keep passing.
- [ ] Full offline unit-test suite in CI.

Cannot run MATLAB locally in this session — please review and validate before merging.

https://claude.ai/code/session_015dKpDXCo8VTS4AHiibNnCp

---
_Generated by [Claude Code](https://claude.ai/code/session_015dKpDXCo8VTS4AHiibNnCp)_


stevevanhooser:
## Closing — the workaround is not retirable, my audit note was wrong

Cloud CI 1 failed on `2ad46c0` with two SyncFiles=true failures on `FileSeriesRoundTripTest`:

- `testMembersSurviveADownloadFromTheCloud` — `DID:SQLITEDB:open`, `chunkdata.bin_1 ... cannot be accessed ... bytes are not on this machine and could not be retrieved`.
- `testTheBatchScopeIsTheSeriesAndNotTheWholeDocument` — same shape.

`testMembersSurviveADownloadFromTheCloudWithSyncFilesFalse` passed on the same commit — the manifest arrives as `ndic://` there and never has a local-only location for this problem to bite on.

## What I missed

I misread what DID-matlab#191 gives us. Reading DID more carefully:

- `fetchSeriesMemberBytes` (sqlitedb.m:1521) iterates **every** location the manifest carries and hands each one, in turn, to the customFileHandler as `sourcePath`, with the member's uid in `ctx.uid`. Remote first, then local. So DID#191 taught DID to *offer* a local-file manifest path to the handler.
- **NDI's `download_file_from_cloud` refuses any sourcePath that doesn't start with `ndic://`** (`didsqlite.m:246`, error `NDI:Didsqlite:UnsupportedFileLocationType`). DID catches that as a miss and moves on.
- The workaround's second `ndicloud` location gave the handler an `ndic://` URL to actually service on member fetches. Without it — post-retirement — the manifest's only location is a local `file` path; the handler refuses; there are no more locations to try; DID reports the member as absent.

So DID#191 is a necessary but not sufficient condition. The retirement is only safe if the handler can service a local sourcePath OR if the manifest is guaranteed to also carry an `ndic://` location. The retirement removed the guarantee.

## What #201 does and doesn't cover

DID#201's fetch is for the **manifest itself** when it's cloud-only. In the SyncFiles=false shape, the manifest starts as `ndicloud`, so the handler gets an `ndic://` URL from the moment the doc lands — no local-file case to worry about. The SyncFiles=false test in NDI-matlab#988 passes for that reason.

The SyncFiles=**true** shape is different: the manifest bytes come down on disk, `updateFileInfoForLocalFiles` reshapes its location to `file://...`, and — without the workaround — the manifest has no other location. That's the case that broke.

## Paths forward

Two real options; both are follow-ups, not this PR:

1. **Keep the workaround (recommended for now).** Simplest and self-contained; the "belt and suspenders" characterization is wrong.
2. **Teach `download_file_from_cloud` to service a local-file sourcePath** — read the manifest locally to learn the member's uid, then fetch the member from the cloud by composing the `ndic://` URL. Requires threading a `cloudDatasetId` into the handler's context (a DID contract change) or looking it up from the document via the session (session state in a handler is ugly).

I'll re-comment on #986 with this analysis so the "workaround retirement is pending" item can be reframed accurately.

Closing this PR without merging. The offline test I removed (`TestDownloadedSeriesManifestKeepsItsCloudLocation`) stays where it is on `main`, and the branch `claude/ndi-986-retire-manifest-workaround` will remain around as a record of the attempt.

---

